PR #18, opened 2014-05-29
Patched version of protobuf

j-rivero:
A second version of pull request #12

### DIFF
--- a/gazebo.rb
+++ b/gazebo.rb
@@ -14,7 +14,7 @@ class Gazebo < Formula
   depends_on 'freeimage'
   depends_on 'libtar'
   depends_on 'ogre'
-  depends_on 'protobuf'
+  depends_on 'osrf/simulation/protobuf'
   depends_on 'protobuf-c'
   depends_on 'qt'
   depends_on 'sdformat'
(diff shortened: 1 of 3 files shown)


LorenzMeier:
This bug still exists, which lets me guess that this patch is not any more present?
https://github.com/wingtra/sitl_gazebo/issues/4

LorenzMeier:
I've got an up to date system:
```bash
~/src/sitl_gazebo/Build (master)$ brew update
brew upgrade protobufUpdated Homebrew from c7589348 to 11d7d135.
~/src/sitl_gazebo/Build (master)$ brew upgrade protobuf
Error: protobuf 2.6.1 already installed
~/src/sitl_gazebo/Build (master)$ brew upgrade gazebo
Error: osrf/simulation/gazebo 1.9.7 already installed
```

scpeters:
We managed to resolve the issue for our regression tests by fixing the environment variables. Then it seemed like the patched protobuf formula was unnecessary, so we redacted it in #46

I'm on leave this week, so I won't have a change to debug for a little while.

PR #41, opened 2015-07-17
Point the Ogre download at a specific mirror.

mikepurvis:
Bypasses sourceforge's new JS-based redirect dingus that breaks Homebrew.

cf. https://github.com/Homebrew/homebrew/issues/41778#issuecomment-122380024


scpeters:
This causes a `brew audit` complaint, but fixing the URL download is more important. Thanks!

PR #42, opened 2015-07-20
Revert "Point the Ogre download at a specific mirror."

scpeters:
Reverts osrf/homebrew-simulation#41 since sourceforge seems to have recovered.

@mikepurvis


mikepurvis:
+1

PR #46, opened 2015-09-12
Remove custom protobuf formula

scpeters:
In response to https://github.com/Homebrew/homebrew/issues/43520, remove our custom protobuf formula and point `gazebo.rb` to the upstream protobuf formula. I believe the patched protobuf formula is no longer necessary anyway.


LorenzMeier:
The true issue is that "gazebo" in Homebrew doesn't actually refer to something useful, but a super outdated version. I got things running with "gazebo6". Would you accept a PR renaming gazebo to gazebo1 so that people installing it have a chance noticing that you're shipping 7 incompatible (out of which half are hopelessly outdated) versions in parallel? I know that different ROS releases need different versions, but long-term having a clean default would be nice.

scpeters:
Sorry for the delay, I was on away last week.

We can use the [Aliases folder](https://github.com/osrf/homebrew-simulation/tree/master/Aliases) to change the target of the `gazebo` formula. Also, notice that `gazebo1` points to the `gazebo` formula currently. Perhaps we could rename `gazebo.rb` to `gazebo1.rb` and make `gazebo` an Alias that always points to the latest officially released version (currently `gazebo6`)?

I'm cc'ing @wjwwood since I think it would break ROS to change the `gazebo` formula to point to a newer version. What if we created aliases like `gazebo-hydro`, `gazebo-indigo`, `gazebo-jade` for the ROS formulae to depend on?

wjwwood:
I'm ok with how ever you decide to rename them, but I would appreciate it you took the time to make sure the default ROS install still worked as well. The aliased formulae for each ROS distribution probably won't work because you cannot have the `gazebo` rosdep key resolve to different names based only on the ROS distro. I've laid out a strategy for how to deal with this by having a brew tap for each ROS distro which are mutually exclusive (e.g. `brew tap ros/indigo` or `brew tap ros/jade` but not both) and each provide a `gazebo-for-ros` (or some similar name) which points to the correct version of gazebo for the ROS distro that you have tapped.

scpeters:
Ok, so the formula name has to be the same? It can't differ based on the ROS distro? I know there's an issue about this somewhere...

wjwwood:
https://github.com/ros/rosdistro/pull/6334#issuecomment-85783237

wjwwood:
https://github.com/ros/rosdistro/issues/8912

PR #70, opened 2016-01-13
sdformat4 4.0.0

osrfbuild:


sdformat4.rb
```diff
@@ -1,9 +1,9 @@
 class Sdformat4 < Formula
   desc "Simulation Description Format"
   homepage "http://sdformat.org"
-  url "http://gazebosim.org/distributions/sdformat/releases/sdformat-4.0.0~pre3.tar.bz2"
-  version "4.0.0~pre3"
-  sha256 "944a187920603929e68bc31fbce6829dcb1097c0ccaa7e9922306384fcb50644"
+  url "http://gazebosim.org/distributions/sdformat/releases/sdformat-4.0.0.tar.bz2"
+  version "4.0.0"
```
scpeters: when we switch from prerelease to an official release, we can delete the `version` tag. I'll do this manually

scpeters: 2047c51

scpeters:
@j-rivero I'd like to merge this and then add bottles separately. How does that sound?

j-rivero:
> @j-rivero I'd like to merge this and then add bottles separately. How does that sound?

ok for me, the status of this PR probably does not have too much to do with the rest.

PR #77, opened 2016-02-02
ogre: doesn't need freetype patch anymore

scpeters:
We are having [failures compiling ogre](http://build.osrfoundation.org/view/gazebo/job/gazebo-ci-pr_any-homebrew-amd64/82/) since the freetype formula was updated.

I think the fix is to remove a patch.

See also https://github.com/homebrew/homebrew/pull/48666

@wjwwood @j-rivero


scpeters:
I'm going to merge this now to see if it fixes our builds.

scpeters:
ogre is building again for us:

* http://build.osrfoundation.org/job/gazebo-ci-pr_any-homebrew-amd64/84/console

j-rivero:
+1, thanks!

PR #78, opened 2016-02-08
ignition-transport 1.0.1

osrfbuild:


scpeters:
@j-rivero I had created an `ignition-transport1` formula. `ignition-transport` currently refers to version 0. Should I make `ignition-transport` point to the latest major version?

j-rivero:
> @j-rivero I had created an ignition-transport1 formula. ignition-transport currently refers to version 0. Should I make ignition-transport point to the latest major version?

I don't have a good answer for this since I don't know well which are the implications in the brew world. We should probably look at what other projects are doing.

scpeters:
We've been hosting versioned formulae, which is fine with me. We can declare dependencies explicitly based on version. The main issue (I think) is that the release script needs to use a `PACKAGE_ALIAS` parameter that matches the name of the homebrew formula. The build that triggered this PR used `ignition-transport`, while the current setup wants it to be `ignition-transport1`. I can easily change it, but I wanted to check with you about how you like to specify the `PACKAGE_ALIAS` variable.

* http://build.osrfoundation.org/view/os_homebrew/job/generic-release-homebrew_pull_request_updater/38/parameters/

(let me know if this comment didn't make any sense at all)

scpeters:
I've updated this PR, so there's only a single ignition-transport formula and also updated the bottle.

j-rivero:
> We've been hosting versioned formulae, which is fine with me

Note that we usually don't set any version of packages that are in 1.x series. That is the reason why ignition-transport has no version in PACKAGE_ALIAS.

PR #89, opened 2016-04-07
gazebo8 8.0.0-20160406-86b8dcf

osrfbuild:


scpeters:
We are leaving this open to test automatic bottle creation.

PR #90, opened 2016-04-07
sdformat4: treat pkg-config as normal dep

scpeters:
`pkg-config` is both a build-time and run-time dependency (for the test), so treat it as a normal dependency.


scpeters:
we'll also use this PR for testing bottle builds

scpeters:
this won't work for testing PR builds because the branch is owned by osrf not osrfbuild

PR #95, opened 2016-04-21
Update broken gist patch urls in ogre formula

adammw:


scpeters:
Thanks for the fix! I hadn't realized they changed the gist url format.

PR #103, opened 2016-06-09
ignition-transport2: new formula

scpeters:


ignition-transport2.rb
```diff
@@ -0,0 +1,35 @@
+class IgnitionTransport2 < Formula
+  desc "Transport middleware for robotics"
+  homepage "http://ignitionrobotics.org"
+  url "https://bitbucket.org/ignitionrobotics/ign-transport/get/dfbb1c1.tar.gz"
```
caguero: Is this URL correct? I see version 1.2.0 in the CMakeLists.txt inside dfbb1c1.tar.gz

scpeters: you're right, it won't have the same package name. I'll change it to nate's branch

caguero:
Is there any way to add an optional dependency or recommendation to a formula? I'm thinking on ign-tools.

ignition-transport2.rb
scpeters: I just updated the commit to download and squashed

scpeters:
ign-tools is currently a required dependency. we can make it recommended (on by default) or optional (off by default)

PR #105, opened 2016-06-17
ignition-math2 2.4.0

osrfbuild:


scpeters:
The [bottle build fails](http://build.osrfoundation.org/view/os_homebrew/job/generic-release-homebrew_bottle_builder/40/) because some sdformat and gazebo formulae can't be installed side-by-side. I believe the recommended fix is to declare conflicting packages `keg_only`, but I don't think that will work for libraries...

scpeters:
added some `conflicts_with` tags that may fix the bottle build

PR #110, opened 2016-07-08
Remove osx10.11 image name as it's no longer supported

solarce:
The `osx_image: osx10.11` image name was not meant to be a long term supported thing, we've transition to just `xcodeX.x` based names, see https://docs.travis-ci.com/user/languages/objective-c/#Supported-Xcode-versions

Any builds on Travis will, unfortunately, fail, as we didn't realize this tag name was in use in the _wild_ anymore, until you merge in this change.

Thanks for understanding.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ matrix:
   include:
     - env: OSX=10.11
       os: osx
-      osx_image: osx10.11
+      osx_image: xcode7.3
       rvm: system
     - env: OSX=10.10
       os: osx


scpeters:
thanks, I was wondering why our builds were hanging

scpeters:
There were some zombie builds that kept restarting themselves and hanging. I imagine that's why the jobs for this build haven't started yet. I just killed them, but I haven't seen any action from this build yet.

scpeters:
there we go

jslee02:
Is there a specific reason why not using the most latest Xcode versions for each target OS X versions (e.g., using Xcode 7 for OS X 10.10 rather than Xcode 7.1)?

solarce:
@jslee02 is your question directed at me? If so, I'm not entirely sure what you're asking about

jslee02:
It was for @scpeters or any maintainer of this repo who knows the reason.

My question: According to the [link](https://docs.travis-ci.com/user/languages/objective-c/#Supported-Xcode-versions) you provided, each OS X version supports several Xcode versions (e.g., OS X 10.10 supports Xcode 6.3, 6.4, 7, and 7.1). I asked if there is a reason why Xcode 7 is picked over others.

solarce:
@jslee02 each Xcode version has a completed separate base VM image, in order to keep things straightforward, so you don't get one 10.10 image with multiple Xcode versions on it. Does that make sense?

jslee02:
That totally makes sense. I should have made my question clearer.

If I'm correct, @scpeters wants to test the homebrew formulas of this repo for different OS X versions instead of for all the Xcode versions. I wondered why he picked OS X 10.10 + Xcode 7 combination instead of OS X 10.10 + Xcode 6.3/6.4/7.1, for example.

scpeters:
It's arbitrary. I mainly wanted one for each osx distro. One could argue to use the most recent xcode version for each distro, but I don't really care to be updating this travis metadata all the time, unless there are bugs with specific versions.

scpeters:
Thanks for the fix!

jslee02:
Ah, I see. Thanks for the clarification. I should update the travis script for DART too as @solarce suggested here.

PR #115, opened 2016-07-18
gazebo8: latest from default branch

scpeters:


scpeters:
waiting for bottle build

Issue #120, opened 2016-07-28
ignition-transport 1.3.0 compilation failure


bowang:
```
Last 15 lines from /Users/bowang/Library/Logs/Homebrew/ignition-transport/02.make:
/tmp/ignition-transport-20160727-16750-r7edtp/ignition-transport-1.3.0/include/ignition/transport/SubscriptionHandler.hh:148:53: error: expected expression
          auto msgPtr = google::protobuf::down_cast<const T*>(&_msg);
                                                    ^
7 errors generated.
make[2]: *** [src/CMakeFiles/ignition-transport1.dir/ign.cc.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [src/CMakeFiles/ignition-transport1.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
[  8%] Linking CXX static library libgtest.a
cd /tmp/ignition-transport-20160727-16750-r7edtp/ignition-transport-1.3.0/test && /usr/local/Cellar/cmake/3.6.1/bin/cmake -P CMakeFiles/gtest.dir/cmake_clean_target.cmake
cd /tmp/ignition-transport-20160727-16750-r7edtp/ignition-transport-1.3.0/test && /usr/local/Cellar/cmake/3.6.1/bin/cmake -E cmake_link_script CMakeFiles/gtest.dir/link.txt --verbose=1
/usr/bin/ar qc libgtest.a  CMakeFiles/gtest.dir/gtest/src/gtest-all.cc.o
/usr/bin/ranlib libgtest.a
[  8%] Built target gtest
make: *** [all] Error 2
```

OS: Mac OSX 10.11.6 El Capitan
Compiler: Apple LLVM version 7.3.0 (clang-703.0.29)

scpeters:
Thanks for the report. I just set up a jenkins build to replicate the failure, since I'm still using 10.10 on my personal machine ([the first build failed](http://build.osrfoundation.org/job/ignition_transport-ci-pr_any-homebrew_elcap-amd64/1/) for an unrelated issue (missing `hg` installation), cc @j-rivero ). We'll continue working to reproduce and diagnose the issue.

uframer:
I encountered the same issue. It looks like the "osrf/simulation/ignition-transport: stable 1.3.0" package depends on protobuf260 but now "protobuf: stable 3.0.2" is the default protobuf version.

[...]

PR #132, opened 2016-09-16
Fix build error of ignition-transport 1.3.0

uframer:
This patch fix the issue
https://github.com/osrf/homebrew-simulation/issues/120.
ignition-transport depends on protobuf260 and the current protobuf
version in Home brew is 3.0.2. So we have to change the dependency
settings to protobuf260.


uframer:
Discard this pull request as j-rivero has already submitted a patch to the upstream repo.

scpeters:
See #133

PR #134, opened 2016-09-16
Introduce the curl dependency on gazebo7 and gazebo8

j-rivero:


scpeters:
I don't think this is necessary. Is there a problem with the system version of curl? Or is this just about the pkg-config issue?

j-rivero:
> I don't think this is necessary. Is there a problem with the system version of curl? Or is this just about the pkg-config issue?

Gazebo was not able to locate the curl package so I assumed that we will need to install it via brew. If there is a way of detecting the system one, that could work yes.

scpeters:
No longer needed since upstream homebrew-core pkg-config formula has been fixed

PR #137, opened 2016-09-23
Ignition msgs fix pkg-config file

scpeters:


j-rivero:
I've remove the restriction of just two bottle files and the [generation job finished just fine](http://build.osrfoundation.org/job/generic-release-homebrew_bottle_builder/80/console). There is only one .bottle.tar.gz for `ignition-msgs` but I assume that this is just fine since `ignition-transport` is not changing the source code only bumping the revision (which affects to build metadata only).

scpeters:
Actually, the `mv` script just missed the second bottle because it's named `*.bottle.1.tar.gz`:

~~~
diff -r 2d724aedcdde jenkins-scripts/lib/homebrew_bottle_creation.bash
--- a/jenkins-scripts/lib/homebrew_bottle_creation.bash	Sat Sep 24 02:16:56 2016 +0200
+++ b/jenkins-scripts/lib/homebrew_bottle_creation.bash	Fri Sep 23 17:51:49 2016 -0700
@@ -36,7 +36,7 @@
   exit -1
 fi
 
-mv *.bottle.tar.gz ${PKG_DIR}
+mv *.bottle*.tar.gz ${PKG_DIR}
 mv *.bottle.json ${PKG_DIR}
 
 echo '# END SECTION'
~~~

scpeters:
The ign-msgs and ign-transport2 bottles are successfully uploaded by the support_multiple_bottles branch, but the pr_bottle_hash_update script needs to be updated too.

http://build.osrfoundation.org/view/os_homebrew/job/generic-release-homebrew_pr_bottle_hash_updater/76/console

scpeters:
The bottles aren't downloading correctly. We should check where they were uploaded

scpeters:
both bottles got uploaded to the wrong folder:

* http://build.osrfoundation.org/view/os_homebrew/job/repository_uploader_ng/2890/console

~~~
+ ./s3cmd put /var/lib/jenkins/jobs/repository_uploader_ng/workspace/pkgs/ignition-msgs-0.5.0_1.yosemite.bottle.tar.gz s3://osrf-distributions/ignition-msgs/releases/
...
+ ./s3cmd put /var/lib/jenkins/jobs/repository_uploader_ng/workspace/pkgs/ignition-transport2-2.0.0_1.yosemite.bottle.tar.gz s3://osrf-distributions/ignition-msgs/releases/
~~~

prefix: `ignition-msgs`, but they should be in `ign-msgs` and `ign-transport`

j-rivero:
> prefix: ignition-msgs, but they should be in ign-msgs and ign-transport

I see. For getting the short prefix (ign instead of ignition) I don't see many options a part of hardcoding the change in the repository_uploader script. Now that we can get more than one bottle per script run, we need to extract destination. Or we do it from the bottle filename or we export .json files together with .bottles files and parse the information from there.

scpeters:
I made a mistake when I started the bottle builder job, putting `ignition-msgs` instead of `ign-msgs` in the `PACKAGE` field:

* http://build.osrfoundation.org/view/os_homebrew/job/generic-release-homebrew_bottle_builder/86/parameters/

but we shouldn't have to rely on that.

I think the logic won't be too hard to determine the prefix:

* take formula name
* remove trailing numbers
* replace `ignition-` with `ign-`

that should be enough

scpeters:
I uploaded these manually because this is breaking gazebo homebrew default builds. We can continue testing with #129

PR #139, opened 2016-09-30
New formulae for ign-common

j-rivero:
Patches for compilation [have been submitted to upstream](https://bitbucket.org/ignitionrobotics/ign-common/pull-requests/13)

### DIFF
--- a/ignition-common.rb
+++ b/ignition-common.rb
@@ -29,6 +29,28 @@ class IgnitionCommon < Formula
   end
 
   test do
-    system "pkg-config", "--modversion", "ignition-common"
+    (testpath/"test.cpp").write <<-EOS.undent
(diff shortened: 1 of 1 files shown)


scpeters:
Can you make the patches download from bitbucket instead of attaching them inline?

For example:

* https://github.com/osrf/homebrew-simulation/blob/master/gazebo4.rb#L50-L54

This way it's easier to track the reasoning and pull request history.

scpeters:
Also, the patch fails to apply to the tarball:

* https://travis-ci.org/osrf/homebrew-simulation/jobs/164059745#L866-L873

scpeters:
I've been trying to clean up this formula, and I added a better test, but I'm running into problems with the location of the installed header files.

PR #150, opened 2016-10-10
sdformat4 4.2.0

osrfbuild:


scpeters:
this formula still had the old revision number; I'm re-running the jenkins job with a test branch that will remove outdated revisions

scpeters:
see #151

PR #159, opened 2016-11-22
Bump revision on gazebo7 due to problems with bullet dependency

j-rivero:
Bullet recently got updated and it changed libraries [name making gazebo binary not able to find previous bullet libraries](http://build.osrfoundation.org/view/All/job/gazebo-install-one_liner-homebrew-amd64/336/consoleFull#-1972803183b8221636-6cd8-440f-bb53-7855dac94d03).

### DIFF
--- a/gazebo7.rb
+++ b/gazebo7.rb
@@ -3,6 +3,7 @@ class Gazebo7 < Formula
   homepage "http://gazebosim.org"
   url "http://gazebosim.org/distributions/gazebo/releases/gazebo-7.4.0.tar.bz2"
   sha256 "a033b2273383f16e5dd5b5bfe597551dc3618b98e64abecfa8a41bdddd6542f7"
+  revision 1
 
   head "https://bitbucket.org/osrf/gazebo", :branch => "gazebo7", :using => :hg
 


scpeters:
The bottle builder job won't run automatically yet. It has to be triggered automatically.

Also, I think the PR updater will fail if the PR doesn't come from the `osrfbuild` account. I'm away right now, but I can advise about how to fix this.

Also, I'd recommend bumping revisions on gazebo5 and gazebo6, since those bottles are probably broken too.

j-rivero:
> Also, I think the PR updater will fail if the PR doesn't come from the osrfbuild account. I'm away right now, but I can advise about how to fix this.

Sure, feel free to propose what you think is the best way to fix it and I will do it.

scpeters:
I would push this branch to the [osrbuild fork](https://github.com/osrfbuild/homebrew-simulation) and re-create the pull request from there. Then manually start a [bottle build in jenkins](http://build.osrfoundation.org/view/os_homebrew/job/generic-release-homebrew_bottle_builder/).

We could potentially simplify this process if we allow `osrfbuild` to push to this repo, but it would need testing.

scpeters:
I pushed this branch to `osrfbuild` in #161

scpeters:
fixed in #161

Issue #170, opened 2016-12-17
bottle for ogre1.9


scpeters:
I forgot to build one in #166

LorenzMeier:
I would propose to re-open and add bottles for Sierra. More and more folks will be on it. What works as a robust work-around is to copy the el-capitan bottle on the server and to add a line for sierra in the brew file.

Issue #160, opened 2016-11-25
Cannot build for MacOS Sierra (10.12.1)


chenny408:
The makefile produces an error when during installation.  Is there a way to install gazebo for this mac version?

brew install gazebo7 

==> Installing gazebo7 from osrf/simulation
qt: This formula either does not compile or function as expected on macOS
versions newer than El Capitan due to an upstream incompatibility.
Error: An unsatisfied requirement failed this build.

scpeters:
We are currently migrating gazebo8 to use qt5, so that version will be work for MacOS Sierra, but older versions of gazebo (7 and under) won't work on MacOS Sierra. Gazebo8 is in development in the `qt5` branch. See [gazebo issue 1804](https://bitbucket.org/osrf/gazebo/issues/1804/migrate-to-qt5) as well for more info.

chenny408:
ok got it, when's the planned release date for Gazebo8?

scpeters:
release scheduled for Jan 25 2017: http://gazebosim.org/#status

starthal:
Refer to PR #198 for Gazebo 7 installation on Sierra.

scpeters:
gazebo7 and gazebo8 now have bottles on sierra

PR #172, opened 2016-12-21
ogre1.9: add placeholder bottle block

scpeters:


scpeters:
fixes #170

LorenzMeier:
Yeah, but I'm on sierra, like many other folks 8). The dirty little trick that works reliably for me is to just copy the line for el capitan for sierra.

scpeters:
that may work for ogre, but gazebo7 and earlier won't work on 10.12 due to a lack of qt4 (#160)

we are migrating gazebo8 to qt5 (still in process) and will release in January

LorenzMeier:
As I said we can help with bottle building locally. I can build a working bottle on my machine for sierra.

scpeters:
If you can test the formulae that can use the same bottles on sierra as el_capitan and open a pull request that bumps the formula revision and adds a new line in the bottle block for sierra (much like #174), then I'll be happy to copy the bottles over and/or upload bottles from your team.

As I mentioned, though, an important gazebo7 dependency is not available on sierra, so hopefully gazebo8 will work for your team.

PR #176, opened 2016-12-29
Fix ogre 1.9 FindOGRE.cmake

iche033:
Fix FindOGRE cmake so that ogre dynamic libs can be found. I'm finally able to compile gazebo plugin examples with this patch.

Problem was that ogre 1.9 assumes that dylibs on OSX are named libOgre*.dylib (which is probably true when it's built and installed as a framework) instead of libOgreMain*lib.

### DIFF
--- a/ogre1.9.rb
+++ b/ogre1.9.rb
@@ -28,7 +28,7 @@ class Ogre19 < Formula
   end
 
   patch do
-    # this is the same patch as hgaiser's `window.patch` above, but applicable to the latest 1.9 version
+    # retain osx cocoa window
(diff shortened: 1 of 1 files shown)


iche033:
I had to comment out the bottle section in the formula to test it.. not sure what needs to be done on the bottle end of things

scpeters:
We need to run the bottle builder job, which currently has to be triggered manually

scpeters:
The bottle builder also requires that the pull request comes from the `osrfbuild` account, so I'm submitting another PR

PR #182, opened 2017-01-04
gazebo7: patch to use qt4-no-webkit

scpeters:


serser:
I see that this package has some upstream problems, and you guys have had a solution to it. How could I walk around to use it?

> qt4-no-webkit: This formula either does not compile or function as expected on macOS
> versions newer than El Capitan due to an upstream incompatibility.
> Error: An unsatisfied requirement failed this build.

scpeters:
I recommend the `cartr/qt4` tap

* https://github.com/cartr/homebrew-qt4/blob/master/qt%404.rb

serser:
Thanks @scpeters, to note it works for gazebo 1 to 8 except gazebo7

starthal:
@serser I have been able to install and run Gazebo 7 on Sierra by editing the `qt4-no-webkit` formula and changing the
```
depends_on MaximumMacOSRequirement => :el_capitan
```
to
```
depends_on MaximumMacOSRequirement => :sierra
```
Maybe this breaks something else, but it lets me use Gazebo for now. YMMV.

scpeters:
I've been working on switching to a qt4 formula hosted by a 3rd party tap in #240, but I'm having some trouble with getting the bottles built using `brew test-bot`.

serser:
@starthal  After `brew edit qt4-no-webkit` as you suggested, I did successfully install `qt4-no-webkit`. However, when it comes to `gazebo7`, it asks to install `ignition-math2` while ` ignition-math3` is already installed, thus causing a collision. I am worried if our twist may cause implicit issues if I keep on muting the warnings and errors.

serser:
@starthal It continues breaking while muting, in the end, I still can not install `gazebo7`. It runs into error while making.

starthal:
@serser I didn't have `ignition-math3` installed. I never tried to have `gazebo7` and `gazebo8` installed simultaneously. But `brew unlink ignition-math3` should do the job.

PR #197, opened 2017-01-16
Use ignition-math3 in ignition-msgs

j-rivero:
According to https://bitbucket.org/ignitionrobotics/ign-msgs/pull-requests/22

### DIFF
--- a/ignition-msgs.rb
+++ b/ignition-msgs.rb
@@ -1,15 +1,16 @@
 class IgnitionMsgs < Formula
   desc "Middleware protobuf messages for robotics"
   homepage "https://bitbucket.org/ignitionrobotics/ign-msgs"
-  url "http://gazebosim.org/distributions/ign-msgs/releases/ignition-msgs-0.6.0.tar.bz2"
-  sha256 "aaac3f5ffdbc0e81afde4aa2c297bec52293cfa4091adde9d6fd539089cfd269"
-  revision 1
+  url "http://gazebosim.org/distributions/ign-msgs/releases/ignition-msgs-0.6.1.tar.bz2"
+  sha256 "b23d613d038b94eae0292f38d8e2f68b59d62a239fb2d62405758f47249baf3b"
+  revision 2
 
   head "https://bitbucket.org/ignitionrobotics/ign-msgs", :branch => "default", :using => :hg
 
   bottle do
     root_url "http://gazebosim.org/distributions/ign-msgs/releases"
-    sha256 "fb18db87fb211d1cf5c169ef7d0fe26d7f0fcd4364eb65c3228d9b3692845c9f" => :yosemite
+    sha256 "eb38300b2ebb3136c413bde3968042d03761f5f39549992984677f155f8b01ec" => :el_capitan
+    sha256 "20bf3b3851521e8815126bfd280e060367a0e21e536a4248658426bef0922611" => :yosemite
   end
 
   depends_on "cmake" => :run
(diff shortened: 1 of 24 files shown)


chapulina:
It looks like Travis isn't happy, should it be?

j-rivero:
> It looks like Travis isn't happy, should it be?

The command `brew audit --online` is failing complaining about the lack of the ignition-math2 library in the system. I wasn't able to find documentation about what is the audit command doing. My hypothesis is that it's running otool to check the binary dependencies on the bottle binary which uses ignition-math2 and it can not find it. I'm not sure about how to confirm it.

@scpeters any idea?

scpeters:
We need to build new bottles. I posted instructions in the internal wiki. Basically, we need to bump the formula revision number, then trigger the bottle build jobs. Though I guess the pull request needs to come from the `osrfbuild` tap... I need to update the wiki.

ignition-msgs.rb
```diff
@@ -3,7 +3,7 @@ class IgnitionMsgs < Formula
   homepage "https://bitbucket.org/ignitionrobotics/ign-msgs"
   url "http://gazebosim.org/distributions/ign-msgs/releases/ignition-msgs-0.6.1.tar.bz2"
   sha256 "b23d613d038b94eae0292f38d8e2f68b59d62a239fb2d62405758f47249baf3b"
-  revision 2
+  revision 4
```
scpeters: this one could be bumped to 3 instead of 4, but it doesn't matter

scpeters:
we probably need to bump the revisions for downstream packages, like gazebo7 and gazebo8

scpeters:
may as well merge #202 in as well

chapulina:
+1

j-rivero:
I really can not understand how the hell works the sha system in brew. Let's review the situation, the build is failing with:

```
==> Downloading http://gazebosim.org/distributions/ign-msgs/releases/ignition-msgs-0.7.0.el_capitan.bottle.tar.gz

Downloaded to: /Users/travis/Library/Caches/Homebrew/ignition-msgs-0.7.0.el_capitan.bottle.tar.gz

SHA256: e08df98a1718f985346379de449672001a060845d68b217116358e8bf7a0313d

==> Retrying download

==> Downloading http://gazebosim.org/distributions/ign-msgs/releases/ignition-msgs-0.7.0.el_capitan.bottle.tar.gz

Warning: Formula reports different SHA256: b2d57d180ac752b27b054013d1ecedef9ed8da10dda43efda6dc380e69ddd1cf

Downloaded to: /Users/travis/Library/Caches/Homebrew/ignition-msgs-0.7.0.el_capitan.bottle.tar.gz

SHA256: b428c2f7e9b95a3d0ebb94f26e3af1350d85b1a1bc704001e0dcd20527557980
```

In that piece on log there are three different hashes for only one bottle. Which one is the good one for brew?

Because if I calculate the hash in my linux system, we have another candidate:
```
jrivero@nium tmp $ shasum -a 256 ignition-msgs-0.7.0.el_capitan.bottle.tar.gz 
a0632f8ae15fdcd2ecf352e1243daf9b42b3509d9d9478b3211eda872528e70e  ignition-msgs-0.7.0.el_capitan.bottle.tar.gz

```
@scpeters SOS

scpeters:
The problem is that isn't not successfully downloading a bottle because the 0.7.0 bottle hasn't been uploaded to the expected location:

* http://gazebosim.org/distributions/ign-msgs/releases/

So whatever it downloads is probably some html with an error about an incorrect address

j-rivero:
> So whatever it downloads is probably some html with an error about an incorrect address

Totally right, with some text that changes every time. I was turning crazy.

j-rivero:
Follow in #212

PR #199, opened 2017-01-17
Use ignition-math3 in sdformat

j-rivero:
According with: https://bitbucket.org/osrf/sdformat/pull-requests/297

### DIFF
--- a/sdformat4.rb
+++ b/sdformat4.rb
@@ -17,7 +17,7 @@ class Sdformat4 < Formula
 
   depends_on "boost"
   depends_on "doxygen"
-  depends_on "ignition-math2"
+  depends_on "ignition-math3"
   depends_on "pkg-config" => :run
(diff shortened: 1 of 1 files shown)


chapulina:
I'm not sure what's the difference between the two Travis jobs, is it expected that the `/pr` one fails?

j-rivero:
Same scenario that the one I described in [the ignition-msgs PR](https://github.com/osrf/homebrew-simulation/pull/197#issuecomment-273548384).

scpeters:
It needs new bottles, as discussed in #197

scpeters:
it will probably be easier too if this was merged into #197, so the bottles could just be built once

chapulina:
This PR should be updated for sdf5, right?

j-rivero:
> This PR should be updated for sdf5, right?

Right. Let's continue in #197 and a new PR for sdformat5

PR #211, opened 2017-01-26
New sdformat5 formula

osrfbuild:


scpeters:
Can you do "Squash and Merge" next time so it keeps the history cleaner?

PR #212, opened 2017-01-26
Use Ign math3 in msgs and gazebo8

osrfbuild:


scpeters:
if you can use "Squash and merge", it will keep the history cleaner

j-rivero:
> if you can use "Squash and merge", it will keep the history cleaner

Do you want to wait for Travis or we should be good enough to go?

scpeters:
looks ok to me, though the framerate is terrible. I showed it to @iche033 and we are investigating

it's safe to merge

iche033:
I was able to reproduce the framerate issue with Qt5.8 and it's fixed by enabling the `Qt::WA_PaintOnScreen` attribute - which is being done in [this pull request](https://bitbucket.org/osrf/gazebo/pull-requests/2562/load-gui-plugins-with-gzclient-parameter-g/diff)

PR #226, opened 2017-04-10
ignition-common 0.2.0

osrfbuild:


scpeters:
The shasum was wrong for some reason, so I fixed that in e3e7aca4dec207ca7e9cf19a1791bc3c911b0f6f

I also added the bottle blocks from #224 in c53d9df36acb93b11130bbb282688849d1269619

scpeters:
So I think we just need to re-trigger the bottle builder jobs

j-rivero:
> The shasum was wrong for some reason, so I fixed that in e3e7aca

That was my fault, the problem came from my problems uploading to S3.

> So I think we just need to re-trigger the bottle builder jobs

Let me try it again.

j-rivero:
we needs to merge first pr #227

scpeters:
I think we should add that commit to this PR, since the patch is needed for version 0.1.0

scpeters:
I just cherry-picked it in 1f9cee0d52915c21b669f50e8aa55cf5d1eb0261

scpeters:
I guess we need an ignition-math 3.1.0 release since ignition-common 0.2.0 needs Color.hh

j-rivero:
> I guess we need an ignition-math 3.1.0 release since ignition-common 0.2.0 needs Color.hh

yeeep!

PR #229, opened 2017-04-18
Added the ignition-rndf formula

j-rivero:
/cc @caguero

### DIFF
--- a/ignition-rndf.rb
+++ b/ignition-rndf.rb
@@ -9,8 +9,8 @@ class IgnitionRndf < Formula
   bottle do
     root_url "http://gazebosim.org/distributions/ign-rndf/releases"
     cellar :any
-    sha256 "todo" => :el_capitan
-    sha256 "todo" => :yosemite
+    sha256 "0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF" => :el_capitan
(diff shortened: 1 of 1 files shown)


scpeters:
There's a few complaints in [travis](https://travis-ci.org/osrf/homebrew-simulation/builds/223264390) from the `brew test-bot`:

~~~
==> brew audit osrf/simulation/ignition-rndf --online --new-formula
Error: 4 problems in 1 formula

osrf/simulation/ignition-rndf:
  * C: 2: col 3: Description is too long. "name: desc" should be less than 80 characters.
    Length is calculated as IgnitionRndf + desc. (currently 89)
  * C: 6: col 1: Trailing whitespace detected.
  * Stable: sha256 should be 64 characters
  * Stable: sha256 contains invalid characters
~~~

ignition-rndf.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  def install
+    mkdir "build" do
+      system "cmake", "..", *std_cmake_args
+      system "make", "install"
+    end
+  end
+
+  test do
+    system "false"
```
scpeters: we should add a test that compiles an example program and runs it. see [sdformat5](https://github.com/osrf/homebrew-simulation/blob/master/sdformat5.rb#L42) for an example

PR #231, opened 2017-04-22
New ignition-rndf formula (take II)

j-rivero:
Replaces pull request #229

### DIFF
--- a/ignition-rndf.rb
+++ b/ignition-rndf.rb
@@ -6,10 +6,12 @@ class IgnitionRndf < Formula
 
   head "https://bitbucket.org/ignitionrobotics/ign-rndf", :branch => "default", :using => :hg
 
+  revision 1
+
   bottle do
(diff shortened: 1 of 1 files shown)


j-rivero:
I expect the test to fail due to [this bug](https://bitbucket.org/ignitionrobotics/ign-rndf/pull-requests/3/fix-pkg-config-include-path/diff) in upstream code.

scpeters:
I added a patch for the commit from bitbucket that fixes the test, and also fixed the test so it could compile. This looks good to me. Do you want to build bottles now or later?

j-rivero:
> I added a patch for the commit from bitbucket that fixes the test, and also fixed the test so it could compile. This looks good to me. Do you want to build bottles now or later?

we can release a new version that adds the pkg-config commit and build from there. So +1 for merging and remove the patch afterwards.

scpeters:
sounds good. this should allow the homebrew CI jobs to get the dependencies installed

PR #234, opened 2017-05-03
gazebo[1-6]: update option names for bullet/dart

scpeters:
Remove `with-shared` from bullet and change `core-only` to `with-core-only` for dartsim4


scpeters:
These formula fail `brew audit` because there are options specified for the bullet and dartsim4 dependencies. I'm going to ignore these failures because these formulae don't have bottles anymore

PR #240, opened 2017-05-09
Use cartr/qt4/qt@4 instead of qt4-no-webkit

scpeters:
There is a 3rd-party homebrew tap that provides bottles of qt4 without webkit, which was my main objection to using it previously.


j-rivero:
`Error: No available formula with the name "cartr/qt4/qt@4" ` is this expected in the travis testing?

scpeters:
This is an upstream problem; I'm making a pull request to brew to fix it

scpeters:
I just submitted https://github.com/Homebrew/brew/issues/2614 which should fix this build

scpeters:
Actually, the proper fix is in https://github.com/Homebrew/homebrew-test-bot/pull/86

scpeters:
I merged the update to gazebo7 with new bottles in #243, so now this can be merged since it is just updating formulae

PR #261, opened 2017-07-19
New formula for ignition-gui

j-rivero:
The audit method is complaining about no stable version. I don't know if we have a way to workaround this restriction or we need a 0.1 version.


scpeters:
For unreleased packages, I just choose a commit off the default branch and give the link to the bitbucket url like in #191. For the fake version name, probably `0.0.0-20170719-checksum` would work.

ignition-gui.rb
```diff
@@ -0,0 +1,60 @@
...
+    (testpath/"test.cpp").write <<-EOS.undent
+      #include <iostream>
+      #include <ignition/gui/Plugin.hh>
+      int main() {
+        ignition::gui::Plugin foo;
+
+        return 0;
+      }
+    EOS
+    system "pkg-config", "ignition-gui0"
```
scpeters: from travis:

~~~
Failed executing: pkg-config ignition-gui0
~~~

j-rivero:
The whole thing needs patching in the pkgconfig file upstream before be able to pass the test. I'm on it.

j-rivero:
umm the build [is failing in jenkins](https://build.osrfoundation.org/view/All/job/generic-release-homebrew_bottle_builder-elcapitan/57/console#console-section-2) but I'm unable to see which exactly is the error. It is passing in my local system. Any hint?

scpeters:
It fails during the test:

~~~
==> brew test osrf/simulation/ignition-gui --verbose
==> FAILED
Testing osrf/simulation/ignition-gui
==> Using the sandbox
/usr/bin/sandbox-exec -f /tmp/homebrew20170722-52024-1k3nmut.sb /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/bin/ruby -W0 -I /usr/local/Homebrew/Library/Homebrew -- /usr/local/Homebrew/Library/Homebrew/test.rb /usr/local/Homebrew/Library/Taps/osrf/homebrew-simulation/ignition-gui.rb --verbose
==> pkg-config ignition-gui
Error: osrf/simulation/ignition-gui: failed
Failed executing: pkg-config ignition-gui
~~~

I'm not sure why, but that's the place I'd look.

j-rivero:
> I'm not sure why, but that's the place I'd look.

Yep, I saw it. I'm unable to reproduce it locally, I assume that we don't have a better error log. Would you recommend a way to reproduce it in the jenkins node or locally?

scpeters:
Nevermind, there was an unlinked formulae in my workspace

scpeters:
Here's the real clue:

~~~
$ pkg-config --cflags ignition-gui
Package Qt5Core was not found in the pkg-config search path.
Perhaps you should add the directory containing `Qt5Core.pc'
to the PKG_CONFIG_PATH environment variable
Package 'Qt5Core', required by 'ignition-gui', not found
~~~

qt5 is a keg-only package, so it's not in the `PKG_CONFIG_PATH` by default

scpeters:
I filed a ticket with homebrew about our failed bottle builds:

* https://github.com/Homebrew/homebrew-test-bot/issues/115

j-rivero:
@scpeters I assume that we need to wait until homebrew merge your PR before generating the missing bottles? right?

scpeters:
The alternative is to install Xcode on the yosemite and sierra machines, which we may want to install anyway if we plan on making an `ogre@1.10` formula, since it sounds like [that will need Xcode](https://bitbucket.org/osrf/gazebo/pull-requests/2764/gui-plugins-on-macos/diff#comment-42800468).

j-rivero:
> The alternative is to install Xcode on the yosemite and sierra machines, which we may want to install anyway if we plan on making an ogre@1.10 formula, since it sounds like that will need Xcode.

I'll do it and relaunch the jobs. Thanks!

scpeters:
I'm rebuilding bottles with a custom branch for homebrew-test-bot using the following PR:

* https://bitbucket.org/osrf/release-tools/pull-requests/432/specificy-branch-of-homebrew-test-bot-when/diff

Please review that PR as it has already helped merge #266 and #269

scpeters:
It looks like this formula has the same problem with `-` characters in the version string that we had with [ignition-common](https://github.com/osrf/homebrew-simulation/pull/262#issuecomment-318160071). We should change them to `_` or `~` and rebuild the bottles.

PR #262, opened 2017-07-20
New snapshot of ignition-common: 0.4.0-20170720-2071c29

j-rivero:
Generate a snapshot of ignition-common that is able to build gazebo_experimental

### DIFF
--- a/ignition-common.rb
+++ b/ignition-common.rb
@@ -12,7 +12,7 @@ class IgnitionCommon < Formula
     root_url "http://gazebosim.org/distributions/ign-common/releases"
     sha256 "d8331ebc0637c02361e93967444abb3253d54e1e90b74b42d6ed84d2b5e94c5c" => :sierra
     sha256 "d452ceab673157a6b6e8bf0a5ed01a45b66453b9e9c93babb7361a108426b0b8" => :el_capitan
-    sha256 "1ab94541bfcc4c16c8ba576a1f511a65cce3c6d34c6a77d59618ff2a2a93fe5f" => :yosemite
+    sha256 "4e739efdeeb59ca08f8f62307dda205dfc9800ac899bd4e73ef07696ac4e7c69" => :yosemite
   end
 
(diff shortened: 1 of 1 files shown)


scpeters:
I just triggered new bottle builds. Two were successful, but there is a configuration problem on the sierra machine. It's complaining about access to mercurial, but it also has an error similar to one I saw on the following build:

* https://build.osrfoundation.org/job/gazebo-ci-default-homebrew-amd64/247/console

j-rivero:
> I just triggered new bottle builds. Two were successful, but there is a configuration problem on the sierra machine. It's complaining about access to mercurial, but it also has an error similar to one I saw on the following build:

There was a weird configuration in the sierra node in Jenkins point to /usr/local/ for the homebrew installation. I just remove it and the things seems to work fine. Let me know if you see the error again.

scpeters:
Thanks for fixing the build machine configuration!

scpeters:
I think there is an error in the upload script. It thinks the package name is `ign-common-` and has uploaded the bottles here:

* http://gazebosim.org/distributions/ign-common-/releases

scpeters:
I think these are the key lines in the uploader script:

* https://bitbucket.org/osrf/release-tools/src/1ef65afd6d6/jenkins-scripts/lib/repository_uploader.bash#repository_uploader.bash-154:155

If we used an `_` instead of `-` in between date and checksum, would that work?

* `0.4.0-20170720_2071c29`

j-rivero:
> If we used an _ instead of - in between date and checksum, would that work?

I changed it to use `~`, same that we use in our Debian packages. I'm going to launch the bottle builders again.

scpeters:
the sierra build failed for some reason, but that caused the subsequent build to fail since mercurial was not installed. I think the following pull request may help:

* https://bitbucket.org/osrf/release-tools/pull-requests/411/try-getting-release-tools-scripts-using/diff

j-rivero:
ready to merge, I think

PR #279, opened 2017-09-26
build more high sierra bottles

scpeters:


scpeters:
travis has a huge backlog, but I just tested the new bottles and they're all working fine

PR #288, opened 2017-10-05
ignition-math4: new formula

scpeters:
I'm trying out the new homebrew convention for versioned formula with the `@` character. It probably breaks some of our CI scripts.


scpeters:
`brew audit` doesn't like having an `@` versioned formula with `conflicts`:

~~~
Error: 1 problem in 1 formula
osrf/simulation/ignition-math@4:
  * C: 22: col 3: Versioned formulae should not use `conflicts_with`.
    Use `keg_only :versioned_formula` instead.
~~~

It's a hassle to have keg_only libraries, so I think I'll continue with our naming scheme for now. I may rename unsupported formula to use the `@` syntax and make them keg_only to see how that goes.

scpeters:
I just pushed a few commits experimenting with `@` naming for old formulae:

* 1426cfc8e74bb3753e76e94cd07687e413c053e7
* 37f7636e140801e6c5a006365a57884cb6b1214a
* 7ee3f1f4f2fe4b09bc48aa27d29ba2ddfca7ac7e

PR #289, opened 2017-10-05
ignition-cmake -> ignition-cmake0, add alias

scpeters:


scpeters:
merging before CI so I can test with #288

PR #292, opened 2017-10-06
ignition-cmake: update with unversioned cmake vars

scpeters:


scpeters:
merging now, will test with ignition-math4 bottle builds

PR #314, opened 2017-11-27
ign-sensors, ign-rendering and sdformat6 formulas

j-rivero:


j-rivero:
Some formating problems but a couple of things to solve prior merge:

- Wait for upstream patches to be merged in ign-rendering (already submitted)
- We will need an sdformat version that does not depend on ign-math3 but ign-math4. Current sdformat6 is an alias to sdformat5.

ignition-rendering.rb
```diff
@@ -0,0 +1,88 @@
...
+
+  head "https://bitbucket.org/ignitionrobotics/ign-rendering", :branch => "default", :using => :hg
+
+  #bottle do
+  #  rebuild 1
+  #  root_url "http://gazebosim.org/distributions/ign-rendering/releases"
+  #  sha256 "a435130d701ebdaa58bd29833a26d53c612ae2163a195b0124ad71f0e54675d8" => :high_sierra
+  #  sha256 "8660bbd506869aebc3c01114af9fa651d4ee67eaa9259748a7fc4204aeb01c08" => :sierra
+  #  sha256 "ab9346c2e57d496cc71275d44f6091c6e0792913ca5cc1c88af97e5f099bb37a" => :el_capitan
+  #end
```
scpeters: we should just delete the bottle blocks for now; we can re-add them when we are ready to build bottles

ignition-rendering.rb
```diff
@@ -0,0 +1,88 @@
+class IgnitionRendering < Formula
+  desc "Common libraries for robotics applications. Rendering Library"
+  homepage "https://bitbucket.org/ignitionrobotics/ign-rendering"
+  url "https://bitbucket.org/ignitionrobotics/ign-rendering/get/ca7e2c8d7755131c9374c99a1df41c6eee415739.tar.gz"
+  version "0.0.0~20170721~35b9bae"
```
scpeters: the end of the version string should be part of the commit, and the date should be the date of the commit

ignition-sensors.rb
```diff
@@ -0,0 +1,155 @@
+class IgnitionSensors < Formula
+  desc "Common libraries for robotics applications. Sensors Library"
+  homepage "https://bitbucket.org/ignitionrobotics/ign-sensors"
+  url "https://bitbucket.org/ignitionrobotics/ign-sensors/get/3abcc7ff6b9b1e81b1acc1119a5038647864fe35.tar.gz"
```
scpeters: instead of adding patches, we can just point the URL to the tarball corresponding to the patch

ignition-sensors.rb
```diff
@@ -0,0 +1,155 @@
...
+  #end
+
+  depends_on "cmake" => :build
+
+  depends_on "ignition-msgs1"
+  depends_on "ignition-common0"
+  depends_on "ignition-math4"
+  depends_on "ignition-transport4"
+  depends_on "ignition-rendering"
+  depends_on "sdformat6"
```
scpeters: agree that we need an sdformat6 formula that depends on ignition-math4

PR #326, opened 2017-12-05
ignition-cmake 0.2.0

osrfbuild:


scpeters:
retrying with a different release-tools branch

PR #329, opened 2017-12-09
gazebo8: bump revision for qt 5.1.0

scpeters:
waiting for https://github.com/Homebrew/homebrew-core/pull/21435


ilovezfs:
@scpeters Homebrew/homebrew-core#21435 is shipped.

scpeters:
@ilovezfs thanks for the ping. Coincidentally, we just tagged a new version of this package so I'll be attempting to build it in #330, though tinyxml2 6.0.0 changed the API, so I need to fix that.

Cheers

PR #345, opened 2018-01-04
Ignition rndf cmake

clalancette:
I'm trying to switch ignition-rndf over to using ignition-cmake in https://bitbucket.org/ignitionrobotics/ign-rndf/pull-requests/6 .  As part of that, we need to make a new Homebrew Formula for ignition-rndf0 (I think), with dependencies on ignition-math4 and ignition-cmake0 (I think).  This PR does exactly that.  I've never done this before, so I have no idea if this actually works, and I don't actually know how to test this, so any thoughts appreciated.


scpeters:
There's a couple things happening here:

* We just want one version of ignition-rndf for each major version. I think we should just rename the ignition-rndf formula to ignition-rndf0 and make an alias for ignition-rndf.

* I think math3 and math4 can co-exist if we remove [this conflicts line](https://github.com/osrf/homebrew-simulation/blob/master/ignition-math4.rb#L22). then we might be able to depend on both math3 and math4 in the formula for CI purposes, and then just depend on math4 once we've merged and released support for ignition-cmake

Does that make sense?

clalancette:
Yes, that makes sense to me.  I'll try out what you suggested and see what happens here.  Thanks!

clalancette:
We are almost passing, but Travis is complaining that the downloaded version of http://gazebosim.org/distributions/ign-rndf/releases/ignition-rndf0-0.1.5.el_capitan.bottle.tar.gz doesn't match the SHA256 sum in the formula.  While that is true, the problem is that every time I download that file, I get a different SHA 256 value, so I'm not sure what to change the formula SHA 256 to.  Any thoughts?

ignition-math4.rb
```diff
@@ -19,7 +19,10 @@ class IgnitionMath4 < Formula
   depends_on "ignition-cmake0"
 
   conflicts_with "ignition-math2", :because => "Symbols collision between the two libraries"
-  conflicts_with "ignition-math3", :because => "Symbols collision between the two libraries"
+  # TODO(clalancette): Temporarily disabling this conflicts_with so that CI passes for RNDF;
+  # once we land the changes to switch ignition-rndf over to ignition-cmake, this can be
+  # re-enabled.
+  #conflicts_with "ignition-math3", :because => "Symbols collision between the two libraries"
```
scpeters: I think this might not be a conflict actually since we implemented inline namespaces in ignition-math4

I think we can just delete the line

ignition-rndf0.rb
```diff
@@ -17,7 +17,9 @@ class IgnitionRndf < Formula
 
   depends_on "cmake" => :build
   depends_on "ignition-math3"
+  depends_on "ignition-math4"
   depends_on "pkg-config" => :run
+  depends_on "ignition-cmake0"
```
scpeters: I like to keep these alphabetized

scpeters:
The bottles are all here:

* http://gazebosim.org/distributions/ign-rndf/releases/

There is no bottle ignition-rndf0-* because they were build with the old formula name ignition-rndf-*.

The problem with renaming formulae is that we need to rebuild the bottles. For now, I would propose disabling bottles while we transition to ignition-cmake. I think this is ok since we don't have any other formulae that depend on ign-rndf at the moment.

Another option is to point the url to a specific bitbucket commit in that pull request (see the [sdformat6 formula](https://github.com/osrf/homebrew-simulation/blob/master/sdformat6.rb#L4) for an example).

ignition-math4.rb
clalancette: Removed in 4f2643e.

ignition-rndf0.rb
clalancette: Alphabetized in 4f2643e.

clalancette:
> The problem with renaming formulae is that we need to rebuild the bottles. For now, I would propose disabling bottles while we transition to ignition-cmake. I think this is ok since we don't have any other formulae that depend on ign-rndf at the moment.
> 
> Another option is to point the url to a specific bitbucket commit in that pull request (see the sdformat6 formula for an example).

So I ended up both disabling bottles, and pointing the URL to the specific bitbucket commit.  That may have been overkill, but I don't quite understand how just point the URL to the specific bitbucket commit would make the bottles work.  Anyway, let me know if I should reenable the bottles, and I can do that.

clalancette:
Travis failed because bitbucket is having backend issues.  I'll retrigger travis once they sort things out.

scpeters:
Looks good to me; thanks!

clalancette:
I think bitbucket is a bit better now.  Closing and re-opening to trigger travis.

PR #348, opened 2018-01-05
Gazebo 10 sym links

nkoenig:


Aliases/gazebo10
```diff
@@ -0,0 +1 @@
+../gazebo8.rb
```
chapulina: I think this should be 9, which needs to be changed to have different dependencies than 8

nkoenig: pull request #349 is need first.

scpeters: I just merged #349

nkoenig: Update to point to gazebo9

scpeters:
I don't see any changes in this branch

PR #352, opened 2018-01-05
Added ign-common2 alias

nkoenig:


scpeters:
I've renamed the ignition-common0 formula to ignition-common1 in #353 since we need a new formula anyway for the new release

scpeters:
Actually I'm creating a new ignition-common1 formula and retaining the ignition-common0 formula

scpeters:
actually, we can still do this if we change the new aliases to ignition-common2 and point to ignition-common1

scpeters:
we'll need to wait for #353 if we want the aliases to point at ignition-common1

PR #354, opened 2018-01-08
ignition-cmake 0.3.0

osrfbuild:


scpeters:
@nkoenig I had to manually update the tarball sha256 since two separate tarballs were uploaded

I can show you a trick for restarting debbuilds without uploading a new tarball

Issue #34, opened 2015-01-01
Gazebo4 compilation fail


kt10aan:
Gazebo4 fails to compile stopping around the 74-75% mark.

The full logs are here:
https://gist.github.com/anonymous/fb02091d83e713e479d9

Again, 2 weeks approximately ago, the formula was building fine.

scpeters:
The failure is caused by an interaction between boost 1.57 and the qt4 moc preprocessor. There is a [gazebo issue](https://bitbucket.org/osrf/gazebo/issue/1399/build-fails-with-boost-157) and some fixes in branches that are waiting for code review.

Thanks for the report!

scpeters:
gazebo4 formula fixed as of https://github.com/osrf/homebrew-simulation/commit/98e265acf34df57a61b7e499c60e142c702fd2ed

PR #358, opened 2018-01-11
Add new ign-cmake dependency to fuel-tools.

caguero:
Ignition Fuel Tools will use Ignition CMake when [#34](https://bitbucket.org/ignitionrobotics/ign-fuel-tools/pull-requests/34) is merged. This pull request adds the new Ignition CMake dependency.

@scpeters , do I need to make more changes here? Disable bottles or update sha's?


scpeters:
ignition-cmake is already an implicit dependency via ignition-math4 and ignition-common1, so we can just merge this without building new bottles.

PR #378, opened 2018-02-11
Use dartsim in homebrew-core instead of dartsim/dart/dartsim6

jslee02:
[`dartsim` is accepted to `homebrew-core`](https://github.com/Homebrew/homebrew-core/pull/23903).


scpeters:
Thanks @jslee02 ! I'm making dartsim a default dependency now and building new gazebo9 bottles with this support

jslee02:
@scpeters That's awesome, thanks!

PR #391, opened 2018-03-28
Msgs2 math5 cmake1

nkoenig:
Use math5 and cmake1 with msgs2


scpeters:
I updated ignition-cmake1 in 291cfc379a02b70708a5ee0c62ef7dcfcff83fe5 with the pkg-config fix, so I think this is good to go

PR #408, opened 2018-05-01
ignition-fuel-tools 1.1.0

osrfbuild:


chapulina:
We're holding on a 1.1.0 release until we have a local cache which supports multiple model versions, this can be declined.

PR #416, opened 2018-05-15
ignition-gui: depend on common2, transport5

scpeters:
companion to https://bitbucket.org/ignitionrobotics/ign-gui/pull-requests/107/update-to-use-newest-dependencies/diff


chapulina:
+1

PR #418, opened 2018-05-23
ignition-common 1.1.1

osrfbuild:


chapulina:
It's not clear to me why it failed, let me know how I can help

scpeters:
It seems like a bug in homebrew. I can reproduce it and I've reported it upstream:

https://github.com/Homebrew/homebrew-test-bot/issues/157

scpeters:
I think I found the homebrew bug and fixed it in https://github.com/Homebrew/homebrew-test-bot/pull/158

for now, I'll run the bottle build jobs against that branch

PR #423, opened 2018-06-04
ignition-msgs2 2.0.0~pre3

osrfbuild:


chapulina:
This looks ok, is it ok?

scpeters:
Yeah, it looks fine to me. We don't have a place in this formula for bottles to be specified, so this is all we need for now.

I think we could start adding bottles, but it's easier if we do them in order. We have cmake1, so the next time we prerelease math5, we should add the bottle stuff to the formula. Then we could do some others.

PR #430, opened 2018-06-14
sdformat8: new formula from gz11 branch

scpeters:


scpeters:
bottle builds failed for sierra and el_capitan, so I removed those entries

PR #435, opened 2018-06-20
bump formula revisions for protobuf 3.6

scpeters:
build bottles when https://github.com/Homebrew/homebrew-core/pull/29239 is merged


ilovezfs:
C++11 across the board for this one. Not sure if you already do that :)

scpeters:
most of our formulae are using c++11, but I haven't ever used `needs :cxx11` or `ENV.cxx11` in our formulae

ilovezfs:
`needs :cxx11` just sets fails_with for some old compilers (should be irrelevant for you) and `ENV.cxx11` is hack when the build system itself isn't already setting -std=c++11

So neither of those should be needed.

ilovezfs:
https://github.com/Homebrew/homebrew-core/pull/29239 is merged.

scpeters:
thanks for the heads up!

PR #437, opened 2018-06-20
sdformat8: remove boost dependency

scpeters:


scpeters:
this bottle doesn't build with xcode9

nkoenig:
@scpeters is this a case of requiring a newer version of macos and xcode?

scpeters:
the bottle build complains that it can't find `<any>`

* https://build.osrfoundation.org/view/os_homebrew/job/generic-release-homebrew_bottle_builder-highsierra/180/console

I just checked, and high sierra does have `<experimental/any>`, but it doesn't have `<experimental/variant>`, which is also required for the new sdformat. So, I think this won't work until mojave

nkoenig:
Thanks, and it looks like mojave will be released in september/october of 2018.

scpeters:
see #496

Issue #380, opened 2018-02-13
Consider using homebrew-core/urdfdom


jslee02:
`urdfdom` is now available in `homebrew-core`, and `ros/deps` tap removed it. Here is the use of `ros/deps/urdfdom` in this repo: https://github.com/osrf/homebrew-simulation/search?utf8=%E2%9C%93&q=urdfdom&type=

Issue #447, opened 2018-07-13
gazebo9 fails because of missing ros/deps formula


TSC21:
Dears,

While taping, and brew installing gazebo9, I get the following:
```
==> Installing gazebo9 from osrf/simulation
==> Tapping ros/deps
Cloning into '/usr/local/Homebrew/Library/Taps/ros/homebrew-deps'...
remote: Counting objects: 11, done.
remote: Compressing objects: 100% (11/11), done.
remote: Total 11 (delta 0), reused 4 (delta 0), pack-reused 0
Unpacking objects: 100% (11/11), done.
Tapped 6 formulae (42 files, 50KB).
Error: No available formula with the name "ros/deps/urdfdom" (dependency of osrf/simulation/gazebo9)
==> Searching for a previously deleted formula (in the last month)...
Warning: ros/deps is shallow clone. To get complete history run:
  git -C "$(brew --repo ros/deps)" fetch --unshallow
```
Is this supposed to happens with the `ros/deps` formulas? Any solution for this?

Thanks!

TSC21:
Probably related: https://github.com/osrf/homebrew-simulation/issues/380

jslee02:
I have the same problem.

> Probably related: #380

That's possible. But it's wired because all the `urdf` dependencies of formula in this repo are optional so it shouldn't be a problem. 🤔

TSC21:
> That's possible. But it's wired because all the urdf dependencies of formula in this repo are optional so it shouldn't be a problem. 

True. Let's wait for @scpeters feedback.

scpeters:
I'm not sure why it's happening just to you, but @jslee02 is right in #380, we should switch to the core urdfdom. See #448

PR #448, opened 2018-07-13
sdformat*: optional depend on core urdfdom

scpeters:


scpeters:
fixes #380 and #447

jslee02:
Ouch, I almost finished the PR. You beat me.

PR #457, opened 2018-07-31
ignition-rendering0: use newer commit

dhood:
Context is in https://bitbucket.org/ignitionrobotics/ign-rendering/issues/28
This is the first time I've done this. I'm copying https://github.com/osrf/homebrew-simulation/pull/450. I think the other commits must get generated automatically somehow.

This version of ignition rendering requires an updated version of ignition cmake: see #458

PR #458, opened 2018-07-31
ignition-cmake1: use newer commit

dhood:
I'm looking for an updated version of ignition cmake for ignition rendering (see #457).

This converts it from using a released version to a particular commit, which might not be appropriate. Please advise if a new release is more appropriate instead

### DIFF
--- a/ignition-cmake1.rb
+++ b/ignition-cmake1.rb
@@ -1,9 +1,9 @@
 class IgnitionCmake1 < Formula
   desc "CMake helper functions for building robotic applications"
   homepage "https://ignitionrobotics.org"
-  url "https://bitbucket.org/ignitionrobotics/ign-cmake/get/a7d35cff6874.tar.gz"
-  version "1.1.0~20180727~a7d35cf"
-  sha256 "651be3aa076c5dca1228ca926161bbfca012596ec9bd142c46812ef2550d952d"
+  url "https://bitbucket.org/ignitionrobotics/ign-cmake/get/9b5a8d7e1a5877bc4c0caee8236d7efaa717dfb1.tar.gz"
+  version "1.1.0~20180731~9b5a8d7"
+  sha256 "e9e6f32d06dcfdffaa3e0afd9220c71c9caa36e0125d39e6dc4b78845b902e4c"
 
   head "https://bitbucket.org/ignitionrobotics/ign-cmake", :branch => "default", :using => :hg
 
--- a/ignition-cmake1.rb
+++ b/ignition-cmake1.rb
@@ -10,7 +10,7 @@ class IgnitionCmake1 < Formula
   bottle do
     root_url "http://gazebosim.org/distributions/ign-cmake/releases"
     cellar :any_skip_relocation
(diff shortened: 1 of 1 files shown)


chapulina:
`ign-rendering0` uses `ign-cmake1`, so I think [this file](https://github.com/osrf/homebrew-simulation/blob/master/ignition-cmake1.rb) should be edited instead, right?

And if you notice, that file uses a specific commit, so I think the approach taken here should be ok. This PR should be a good reference: https://github.com/osrf/homebrew-simulation/pull/393

dhood:
thanks, I had looked at the log of [this build](https://build.osrfoundation.org/job/ignition_gui-ci-pr_any-homebrew-amd64/161/console) that referenced `ignition-cmake0`, but not closely enough to realise that's what it's _uninstalling_ :laughing: 
I'll swap to ignition-cmake1 as you suggested!

chapulina:
Thanks for the change, it looks correct to me, but I don't know why Travis is failing... @scpeters , any ideas?

j-rivero:
```
Downloading http://gazebosim.org/distributions/ign-cmake/releases/ignition-cmake1-1.1.0~20180731~9b5a8d7.el_capitan.bottle.tar.gz
curl: (22) The requested URL returned error: 406 Not Acceptable
==> Retrying download
==> Downloading http://gazebosim.org/distributions/ign-cmake/releases/ignition-cmake1-1.1.0~20180731~9b5a8d7.el_capitan.bottle.tar.gz
curl: (22) The requested URL returned error: 406 Not Acceptable
Error: Failed to download resource "ignition-cmake1"
Download failed: http://gazebosim.org/distributions/ign-cmake/releases/ignition-cmake1-1.1.0~20180731~9b5a8d7.el_capitan.bottle.tar.gz
==> FAILED
```
We are changing the version in this PR without touching the bottle information so the script is trying to locate the bottles with the new version and thus failing (they don't exist). We need one of both: or use release.py to release a new version for all our .debs and bottles (instead of this PR) or use Jenkins to build bottles using this PR (non affecting .debs). For this latest one, checkout the internal wiki information under the topic "Building bottles".

scpeters:
Travis is failing because it expects the bottle to be updated. ign-cmake is a special case in that it only needs to build one bottles (for el_capitan) and then it can use it for the other OS versions.

So I recommend starting a bottle builder job for el capitan and passing this pull request URL. I'll do this tomorrow if no one else has done it before then.

* https://build.osrfoundation.org/view/os_homebrew/job/generic-release-homebrew_bottle_builder-elcapitan/

dhood:
Thanks for the pointers! I didn't realise there was an internal wiki page about this: I squashed the commits in this PR as it suggested. I'm having login issues that I'm trying to sort out for the buildfarm to kick off the jobs

dhood:
I got it to commit. But since 1.1.0 is released perhaps I was supposed to use version-0.00000001 like is done [here](https://github.com/osrf/homebrew-simulation/pull/450/files#diff-b43d236cccd0c3d7a7e8ec301d1e2237L5). Please review carefully!

scpeters:
it looks great to me. thanks!

PR #472, opened 2018-08-24
ignition-cmake1 1.1.0~pre3

osrfbuild:


scpeters:
CI complains that the version number has decreased, but I think it's fine

PR #482, opened 2018-09-10
ignition-rendering 1.0.0~pre1

osrfbuild:


scpeters:
release is still being sorted out

PR #488, opened 2018-09-11
Update plugin to use ign-cmak2

nkoenig:
I think I also need to update hashes....


scpeters:
If we want to build new bottles, we need to point to a new tarball (probably from a specific commit), but we can just wait for the next release, since we just have the Factory pull request remaining.

If you just want to fix the homebrew CI, I would just add both `cmake1` and `cmake2` as side-by-side dependencies, and we can address the bottles later.

scpeters:
building against this PR: [Build Status](https://build.osrfoundation.org/job/ignition_plugin-ci-pr_any-homebrew-amd64/38/) https://build.osrfoundation.org/job/ignition_plugin-ci-pr_any-homebrew-amd64/38/

PR #495, opened 2018-09-14
ignition-rendering 1.0.0~pre3

osrfbuild:


scpeters:
merging this now to get the alias fixes, will need another prerelease I believe before bottles can be built

PR #501, opened 2018-09-24
Add ignition-gui@1 alias

j-rivero:
Needed for gz11 branch: https://build.osrfoundation.org/job/ignition_gui-ci-gz11-homebrew-amd64/1/console

### DIFF
--- a/Aliases/ignition-gui1
+++ b/Aliases/ignition-gui1
@@ -1,0 +1,1 @@
+../ignition-gui0.rb
--- a/Aliases/ignition-gui@1
+++ b/Aliases/ignition-gui@1
@@ -1,0 +1,1 @@
+../ignition-gui0.rb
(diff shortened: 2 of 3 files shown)


j-rivero:
> I added a `ignition-gui1` alias also (without the `@`), since I think that is the one that is needed

Oh, what is the difference between both?

> [Build Status](https://build.osrfoundation.org/job/ignition_gui-ci-gz11-homebrew-amd64/2/) https://build.osrfoundation.org/job/ignition_gui-ci-gz11-homebrew-amd64/2/

Failing due to the lack of ign-cmake2. Not sure if we need to change it in all ign-gui formulas or create a real one for ign-gui1 and change the ign-cmake version on that.

scpeters:
One workaround is to declare multiple dependencies if they don't conflict with each other (for example, ign-physics currently [depends on ign-cmake1 and ign-cmake2](https://github.com/osrf/homebrew-simulation/blob/master/ignition-physics0.rb#L14-L15)). This allows the CI to pass, since we still have the default branch distinct from the gz11 branch. We should eventually create a separate formula.

PR #537, opened 2018-11-28
ignition-gazebo0: new formula

scpeters:


scpeters:
the test doesn't pass yet, so we'll build bottles once that is fixed

PR #548, opened 2018-12-21
ignition-fuel-tools2 2.0.0

osrfbuild:


scpeters:
this is modifying the `ignition-fuel-tools1` formula; we should create a separate `ignition-fuel-tools2` formula instead

PR #552, opened 2018-12-27
[WIP] Add ign-fuel-tools3 and ign-gazebo depends on it

chapulina:
Needed by: https://bitbucket.org/ignitionrobotics/ign-gazebo/pull-requests/112

Needs this merged and tagged: https://bitbucket.org/ignitionrobotics/ign-fuel-tools/pull-requests/82


Formula/ignition-fuel-tools3.rb
```diff
@@ -0,0 +1,63 @@
+class IgnitionFuelTools3 < Formula
+  desc "Tools for using fuel API to download robot models"
+  homepage "https://ignitionrobotics.org"
+  url "http://gazebosim.org/distributions/ign-fuel-tools/releases/ignition-fuel-tools3-3.0.0~pre1.tar.bz2"
```
scpeters: @chapulina instead of waiting for a tagged release, I would just point to a bitbucket commit, like in https://github.com/osrf/homebrew-simulation/blob/master/Formula/ignition-gazebo0.rb#L4

the url will be replaced when a  release is made

chapulina: Thanks for tackling it!

scpeters: 972629a0bc4a82a867f832d736697f428358b01f

PR #564, opened 2019-01-15
Update ign-sensors commit

chapulina:


scpeters:
I think this bottle will fail to build until we resolve #565

chapulina:
> I think this bottle will fail to build until we resolve #565

Do you think this will take a while? Otherwise we may merge [this PR](https://bitbucket.org/ignitionrobotics/ign-gazebo/pull-requests/114/wip-camera-sensors/diff) before fixing the Homebrew build.

scpeters:
yeah, don't wait on homebrew; this could take a couple days

PR #595, opened 2019-02-22
ignition-launch0 initial version

j-rivero:
Since we lack of any prerelease, not sure what to set as url for the code. @scpeters I think that we need to wait until merging the current PRs in launch and select the newest commit in the repo, is this right?


scpeters:
I'm not building bottles for now, just merging so the dependency information will be available for CI

PR #605, opened 2019-03-01
sdformat8 8.0.0

osrfbuild:


j-rivero:
@scpeters not sure why only the mojave bottle has been created but seems like the same behavior found during prereleases pull #600

scpeters:
sdformat8 doesn't compile on earlier OS versions because they don't have enough c++17 features, so it's only mojave for sdformat8, physics, gazebo and a few others

PR #608, opened 2019-03-01
ignition-gui 1.0.0

osrfbuild:


scpeters:
the azure pipeline didn't run, but it's probably fine

PR #610, opened 2019-03-01
ign-gazebo0 use ign-sensors1

chapulina:


scpeters:
this should be handled by #606; hopefully it builds properly

chapulina:
I hadn't realized this is already included in https://github.com/osrf/homebrew-simulation/pull/606/

PR #613, opened 2019-03-02
ignition-gazebo 1.0.0

osrfbuild:


scpeters:
the formula needs to be renamed to `ignition-gazebo1`

PR #623, opened 2019-03-06
ignition-gui0 0.1.0

osrfbuild:


scpeters:
the test is failing; I've been working on it locally

also, `Qt5Core` is missing from the pkg-config requires. This should be submitted upstream and patched in this formula

scpeters:
patch submitted upstream:

* https://bitbucket.org/ignitionrobotics/ign-gui/pull-requests/194/use-ign_find_package-for-qt5-and-specify/diff

PR #631, opened 2019-03-12
Alias ignition-gui2 -> ignition-gui1

chapulina:
Goes with this PR: https://bitbucket.org/ignitionrobotics/ign-gui/pull-requests/198/bump-default-to-200/diff


scpeters:
sorry I missed this and already pushed this change in #633

PR #632, opened 2019-03-13
ignition-sensors2, ignition-rendering2 initial versions

iche033:
New ignition-sensors2 formula based on a commit. 

I'm not very sure about the changes. Let me know if I need to change anything. 

I copied from `ignition-sensors1.rb` but removed the `bottle` section and the `version` field. The `url` now points to a specific commit and I also bumped ign-rendering dependency to ignition-rendering2

### DIFF
--- a/Aliases/ignition-rendering2
+++ b/Aliases/ignition-rendering2
@@ -1,1 +1,0 @@
-../Formula/ignition-rendering1.rb
--- a/Aliases/ignition-sensors2
+++ b/Aliases/ignition-sensors2
@@ -1,1 +1,0 @@
-../Formula/ignition-sensors1.rb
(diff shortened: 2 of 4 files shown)


scpeters:
thanks; this formula looks great, but I think it will need a real ignition-rendering2 formula as well since the `find_package(ignition-rendering2)` calls will fail if we are still installing ignition-rendering1 via the alias

iche033:
got it, should I add the ignition-rendering2 formula in the same PR or a separate one?

scpeters:
you can add it here

iche033:
ok added `ignition-rendering2.rb` file in b45e917

scpeters:
thanks @iche033 ! the formulae look good; I merged now so the upstream CI jobs can find the new dependency information. I'll build bottles in a separate pull request

PR #641, opened 2019-03-18
ignition-launch 0.2.0

osrfbuild:


j-rivero:
@scpeters I needed to disable the test since the mojave build was failing. We'll need to inspect that later.

PR #644, opened 2019-03-20
Add ignition-acropolis1 alias

j-rivero:
It will make things easier in the CI since it is currently failing here with our generic script: https://build.osrfoundation.org/job/ignition_acropolis-ci-default-homebrew-amd64/4/


scpeters:
thanks!

PR #645, opened 2019-03-20
Enable the use of ogre2.1 in ignition-rendering

j-rivero:
Add support for ogre2.1 in ignition rendering by adding it as a dependency. Not sure if we want the same for previous majors of ign-rendering, @iche033 ?

### DIFF
--- a/Formula/ignition-rendering2.rb
+++ b/Formula/ignition-rendering2.rb
@@ -21,6 +21,7 @@ class IgnitionRendering2 < Formula
   depends_on "ignition-plugin1"
   depends_on :macos => :high_sierra # c++17
   depends_on "ogre1.9"
+  depends_on "ogre2.1"
 
   def install
     system "cmake", ".", *std_cmake_args


scpeters:
I was never able to get a mojave bottle of ogre2.1 to work. brew allows older bottles to be installed, but ign-rendering fails to compile against it due to use of openGL libraries that aren't supported on macOS 10.14

also, the ogre formulae conflict, since they both install their headers to the same place. we would have to patch the formula similar to how the debs are patched to allow side-by-side installation

cc @iche033 I could use some help trying to compile a version of ogre2.1 that uses metal instead of openGL so it will work on 10.14

j-rivero:
> also, the ogre formulae conflict, since they both install their headers to the same place. we would have to patch the formula similar to how the debs are patched to allow side-by-side installation

I will look into modify the formula to get this

iche033:
> Not sure if we want the same for previous majors of ign-rendering

ign-rendering1 also supports ogre 2.1 so we could try add the dependency there too. 

> I could use some help trying to compile a version of ogre2.1 that uses metal instead of openGL so it will work on 10.14

Actually the ogre2 implementation in ign-rendering depends on OpenGL3+ and it's probably not compatible with Metal. It's something I should probably start paying attention to soon so we can be ready for 10.15

iche033:
where can I find the ign-rendering compile errors against ogre 2.1?

scpeters:
Officially Apple says that OpenGL is only deprecated on mojave, but I'm seeing compilation errors like https://github.com/OGRECave/ogre/issues/817 when I try to build ogre 2.1. I made a comment in that issue to see if we can make Ogre work on 10.14, but it's unclear if that will happen.

* https://github.com/OGRECave/ogre/issues/817#issuecomment-476428196

In the meantime, I have disabled the GL3Plus render engine in #654

scpeters:
> where can I find the ign-rendering compile errors against ogre 2.1?

I made https://github.com/scpeters/homebrew-simulation/pull/3 for debugging purposes to use only ogre2.1 with ignition-rendering2 and started a jenkins job that should use that pull request:

* [Build Status](https://build.osrfoundation.org/job/ignition_rendering-ci-pr_any-homebrew-amd64/226/) https://build.osrfoundation.org/job/ignition_rendering-ci-pr_any-homebrew-amd64/226/

scpeters:
nope, try this one

* [Build Status](https://build.osrfoundation.org/job/ignition_rendering-ci-pr_any-homebrew-amd64/227/) https://build.osrfoundation.org/job/ignition_rendering-ci-pr_any-homebrew-amd64/227/

scpeters:
:cry: let's try this one

[Build Status](https://build.osrfoundation.org/job/ignition_rendering-ci-pr_any-homebrew-amd64/228/) https://build.osrfoundation.org/job/ignition_rendering-ci-pr_any-homebrew-amd64/228/

scpeters:
well build 228 failed to configure with the default branch of ign-rendering; I think I saw a different failure with ign-rendering1, testing here:

[Build Status](https://build.osrfoundation.org/job/ignition_rendering-ci-pr_any-homebrew-amd64/229/) https://build.osrfoundation.org/job/ignition_rendering-ci-pr_any-homebrew-amd64/229/

iche033:
https://build.osrfoundation.org/job/ignition_rendering-ci-pr_any-homebrew-amd64/229/

The error there is useful. We're including GL specific headers from ogre (needed for ign-gui integration). Since the ogre GL3+ plugin is disabled I suspect those headers are not installed. I'll need to address this when migrating to Metal.

scpeters:
I saw different failures on my local machine; I'll try to reproduce and upload a gist

scpeters:
@iche033 here's the other build error I have been seeing when trying to build ign-rendering1 with ogre2.1:

* https://build.osrfoundation.org/view/os_homebrew/job/ignition_rendering-ci-pr_any-homebrew-amd64/230/console

~~~
In file included from ogre2/src/Ogre2Conversions.cc:17:
In file included from ogre2/include/ignition/rendering/ogre2/Ogre2Conversions.hh:27:
In file included from ogre2/include/ignition/rendering/ogre2/Ogre2Includes.hh:70:
/usr/local/include/OGRE/Hlms/Unlit/OgreHlmsUnlit.h:32:10:
 fatal error: 'OgreHlmsBufferManager.h' file not found
#include "OgreHlmsBufferManager.h"
         ^~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
~~~

I see that file in `/usr/local/include/OGRE/Hlms/Common/OgreHlmsBufferManager.h`, but it has trouble with the relative paths.

iche033:
on ubuntu the header file is installed to the components' include dirs so it's able to find it: 
```
$ dpkg-query -L libogre-2.1-dev | grep OgreHlmsBufferManager
/usr/include/OGRE-2.1/Hlms/Pbs/OgreHlmsBufferManager.h
/usr/include/OGRE-2.1/Hlms/Unlit/OgreHlmsBufferManager.h
```

scpeters:
It looks like the header file issue was caused by a recent change to the v2-1 branch

* https://bitbucket.org/sinbad/ogre/commits/2f83a9179c9f9fe086b488386ef7531d84b184de
* https://forums.ogre3d.org/viewtopic.php?f=25&t=94851&p=544455#p544455

maybe we can just revert that patch?

scpeters:
ign-rendering already uses ogre2.1 as of #665

PR #654, opened 2019-03-26
ogre2.1: new commit, fix build for xcode 10+

scpeters:
Officially Apple says that OpenGL is only deprecated on mojave, but I'm seeing compilation errors like https://github.com/OGRECave/ogre/issues/817 when I try to build ogre 2.1. A workaround suggested in that issue is to disable the GL3Plus rendersystem, which I have added here when xcode 10+ is detected. So I think GL3Plus will be built for Sierra, but it will only have Metal for 10.13+.

There is a possibility that we could fix GL3Plus for Xcode10, but I'm not sure how easy that will be, and 10.15 will only support Metal anyway.

I also added a dependency on rapidjson, which is needed by the Hlms material system.

### DIFF
--- a/Formula/ogre2.1.rb
+++ b/Formula/ogre2.1.rb
@@ -15,6 +15,7 @@ class Ogre21 < Formula
   end
 
   depends_on "cmake" => :build
+  depends_on "pkg-config" => :test
   depends_on "doxygen"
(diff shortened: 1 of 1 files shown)


scpeters:
@iche033 this is finally working and I'm going to merge

PR #665, opened 2019-04-04
ogre2.1: allow side-by-side install

scpeters:
ignition-rendering[12]: depend on ogre2.1 also now

### DIFF
--- a/Formula/ogre2.1.rb
+++ b/Formula/ogre2.1.rb
@@ -36,7 +36,7 @@ class Ogre21 < Formula
 
   patch do
     # fix GL3+ compilation with Xcode 10
-    url "https://bitbucket.org/scpeters/ogre-1/commits/14b5dc7fc2d8e1281140d027e1effb4d8a317895/raw"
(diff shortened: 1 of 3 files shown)


Formula/ogre2.1.rb
```diff
@@ -4,6 +4,7 @@ class Ogre21 < Formula
   url "https://bitbucket.org/sinbad/ogre/get/06a386fa64e79a7204a90faf53da1735743f6c2e.tar.bz2"
   version "2.0.99999~pre0~0~20180616~06a386f"
   sha256 "d2e28bfcfbb1277355047c1d8bcd141b05b83af52d277725168e4281eac92a6d"
+  revision 1
```
j-rivero: I think that this change is making the CI to fail

scpeters: that makes travis and azure fail because the bottle hasn't been built yet

once a bottle is built the CI should be happy

scpeters: I'm just having troubles getting a bottle to build successfully

PR #671, opened 2019-04-10
Updating deps for blueprint

nkoenig:


scpeters:
we addressed much of this in #673 and this now has conflicts

PR #680, opened 2019-04-19
ign-gazebo/physics: use google-benchmark

mjcarroll:
Ref: https://formulae.brew.sh/formula/google-benchmark


mjcarroll:
I have since made the `benchmark` dependency optional, so this is no critical to get in until we want to evaluate benchmarks on macOS.

scpeters:
Ok, I'll just merge this without updating the bottles after making sure that `brew audit` passes

PR #681, opened 2019-04-19
Update ign-common3 hash

nkoenig:


Formula/ignition-common3.rb
```diff
@@ -1,9 +1,9 @@
 class IgnitionCommon3 < Formula
   desc "Common libraries for robotics applications"
   homepage "https://bitbucket.org/ignitionrobotics/ign-common"
-  url "https://bitbucket.org/ignitionrobotics/ign-common/get/4191ec638c9a8bb06295651f9564df2a403029aa.tar.bz2"
```
chapulina: It looks like the previous commit (4191...) [already has](https://bitbucket.org/ignitionrobotics/ign-common/src/4191ec638c9a8bb06295651f9564df2a403029aa/src/Filesystem.cc) the `unique*Path` functions, so we should be ok without this PR for [ign-gazebo PR 187](https://bitbucket.org/ignitionrobotics/ign-common/pull-requests/187/filesystem-functions-to-create-unique/diff)

scpeters: we can still update it for battery support, but I'd like to include as well https://bitbucket.org/ignitionrobotics/ign-common/pull-requests/189/profiler-give-hint-on-osx-to-find/diff first

nkoenig: Sounds good. I'll close this PR.

PR #691, opened 2019-04-28
sdformat8: updating hash

nkoenig:


scpeters:
the version and sha256 were changed, but not the url, so I updated that too

PR #700, opened 2019-05-08
ignition-rendering2 blueprint bump

nkoenig:


nkoenig:
I think this is good to go?

scpeters:
I often let the travis and azure CI run after the bottle commit because they check that the bottle is uploaded properly.

scpeters:
the bottles install correctly but `brew test ignition-rendering2` fails because it wants a GPU, so this is safe to merge

PR #701, opened 2019-05-08
Bumpt sensors2 version

nkoenig:


scpeters:
I prefer squash and merge over using a merge commit because it makes the history easier to read

PR #702, opened 2019-05-08
ignition-blueprint: new formula

scpeters:


scpeters:
I manually tested this and ignition-launch1 doesn't have a bottle and fails to compile; I'll try updating that too

caguero:
Travis complains about the missing `test do ... end` block. Is this needed?

scpeters:
ignition-launch1 failed to compile, so I tried fixing that, then I got sucked in to trying to fix the test, but I couldn't get it working since it depends on `ign gazebo` working on macOS, which we couldn't get working the last time we tried:

* https://bitbucket.org/ignitionrobotics/ign-gazebo/pull-requests/296/ign-tool/diff#comment-100904521

I'll probably just fix ignition-launch1 and leave it without a test

PR #818, opened 2019-08-23
ignition-gazebo2 2.8.0~pre1

osrfbuild:


scpeters:
stable release will follow shortly

PR #821, opened 2019-08-24
ignition-msgs3 and dependents: bump revision for protobuf 3.9

scpeters:


scpeters:
travis job failed due to timeout, but other jobs passed, so it should be fine to merge

PR #851, opened 2019-10-11
Make ignition-tools to depend on ignition-cmake2

j-rivero:
According to pull request https://bitbucket.org/ignitionrobotics/ign-tools/pull-requests/6/updating-to-use-ignition-cmake-2/


scpeters:
that pull request changes the major version to 1; I'll make a new copy of the formula for ignition-tools1

j-rivero:
ugh you are right

PR #865, opened 2019-11-07
gazebo7: don't depend on ignition-transport

scpeters:
There's a weird bug that causes the `ignition-tools` bottle builds to fail due to gazebo7 installing `qt@4` and a workaround for this specific case is to remove the `ignition-transport` dependency since it is optional to gazebo7.


scpeters:
[Build Status](https://build.osrfoundation.org/job/generic-release-homebrew_bottle_builder/505/) https://build.osrfoundation.org/job/generic-release-homebrew_bottle_builder/505/

PR #876, opened 2019-11-16
ignition-rendering3 3.0.0~pre2

osrfbuild:


scpeters:
this formula doesn't work on 10.13 anymore, which is fine; I'll just update the formula

~~~
src/Visual_TEST.cc:320:23:
 error: call to unavailable function 'get':
 introduced in macOS 10.14
  EXPECT_EQ(intValue, std::get<int>(value));
                      ^~~~~~~~~~~~~
~~~

scpeters:
trying again:

* [Build Status](https://build.osrfoundation.org/job/generic-release-homebrew_bottle_builder/524/) https://build.osrfoundation.org/job/generic-release-homebrew_bottle_builder/524/

PR #879, opened 2019-11-19
ignition-tools 1.0.0~pre3

osrfbuild:


chapulina:
I think we need to merge this to fix the tests that started failing with the `0.3.0` release.

PR #894, opened 2019-12-01
gazebo7: revision bump for protobuf 3.11

scpeters:
Also add a patch to fix linking with gts, since this has been causing builds to fail:

* [Build Status](https://build.osrfoundation.org/view/main/view/BuildCopFail/job/gazebo-ci-gazebo7-homebrew-amd64/269/) https://build.osrfoundation.org/view/main/view/BuildCopFail/job/gazebo-ci-gazebo7-homebrew-amd64/269/


scpeters:
The bottle test failed due to an issue with the `qt@4` formula we are using from another tap. It is past its end-of-life, and depends on openssl 1.0, which has been removed from homebrew-core since it is approaching its end-of-life. The removal of openssl broke the `qt@4` bottles ( https://github.com/cartr/homebrew-qt4/issues/67 ), but maybe we should consider not supporting gazebo7 any longer on macOS since several dependencies are no longer supported by homebrew.

scpeters:
I've disabled the bottle since the dependencies are not supported

PR #896, opened 2019-12-02
ignition-fuel-tools4 initial version using msgs5

nkoenig:


scpeters:
[Build Status](https://build.osrfoundation.org/job/generic-release-homebrew_bottle_builder/558/) https://build.osrfoundation.org/job/generic-release-homebrew_bottle_builder/558/

j-rivero:
I think we need https://bitbucket.org/ignitionrobotics/ign-msgs/pull-requests/157 and create a new msgs5 prerelease before doing fuel-tools4.

scpeters:
we could also use an older commit of ign-fuel-tools default branch but make sure the formula still has a dependency on `ignition-msgs5`

PR #900, opened 2019-12-03
ignition-gui3 3.0.0~pre1

osrfbuild:


scpeters:
the ignition-gui tests seem a bit flaky, or maybe it's just the ultron CI node. I disabled that node from bottle builds and restarted the bottle builder to see if that fixes it

PR #932, opened 2019-12-11
ignition-citadel 1.0.0

osrfbuild:

PR #933, opened 2019-12-11
ignition-citadel 1.0.0 (take2)

j-rivero:
re-work pull request #932


j-rivero:
I can not find other problem than Xcode version in Azure services making brew doctor to fail. I think is ok to merge.

PR #984, opened 2020-03-16
ign-launch3: new formula

chapulina:
Needed by this PR: https://bitbucket.org/ignitionrobotics/ign-launch/pull-requests/64/


scpeters:
let's just merge this without building a bottle, since it will complain about a new formula not having a test. it should build a bottle the next time this formula is updated

PR #995, opened 2020-04-06
ignition-cmake2 2.2.0

osrfbuild:


scpeters:
the bottle build is failing due to an issue with fuel-tools. from testing locally on my high sierra machine, I see the following error:

~~~
$ pkg-config --libs ignition-fuel_tools3
Package libcurl was not found in the pkg-config search path.
~~~

I believe this is caused by https://github.com/Homebrew/brew/issues/5068. brew supplies `.pc` files for system libraries included in macOS, but they recently disabled them for external software in https://github.com/Homebrew/homebrew-core/pull/52565. There is discussion in https://github.com/Homebrew/brew/issues/5068 about how to fix it.

But I don't know why curl is even in the pkg-config requirements list; it should be in Requires.private

Bo98:
Do you not build in the superenv? It should still be available there.

Either way, the changes outside of the Homebrew superenv have just been reverted for now.

scpeters:
> Do you not build in the superenv? It should still be available there.

the bottle is built in the superenv correctly, but part of our [test block](https://github.com/osrf/homebrew-simulation/blob/master/Formula/ignition-fuel-tools3.rb#L31-L49) is to invoke `pkg-config`, and that was failing

> Either way, the changes outside of the Homebrew superenv have just been reverted for now.

thanks!

scpeters:
I've added a patch from [ign-cmake pr 184](https://bitbucket.org/ignitionrobotics/ign-cmake/pull-requests/184/ensure-private-dependencies-are-encoded-in/diff) and bumped revisions for fuel-tools3/4 to fix their pkg-config files

* [Build Status](https://build.osrfoundation.org/job/generic-release-homebrew_bottle_builder/750/) https://build.osrfoundation.org/job/generic-release-homebrew_bottle_builder/750/

scpeters:
[Build Status](https://build.osrfoundation.org/job/generic-release-homebrew_bottle_builder/751/) https://build.osrfoundation.org/job/generic-release-homebrew_bottle_builder/751/

chapulina:
`brew test` failed again for fuel-tools 3 and 4

scpeters:
yeah I see that. I've adjusted the pkg-config test in a264949 to see if it will give more explanation of why it failed

scpeters:
the patch doesn't help; it still needs to find the `.pc` file even if it is private

they have reverted the changes that broke this to the extent that manually invoking `pkg-config` from a terminal is able to find `libcurl`, but invoking it from a homebrew test (which [uses homebrew's Stdenv](https://github.com/Homebrew/brew/blob/2.1.15/Library/Homebrew/test.rb#L19) and Stdenv [explicitly overrides `PKG_CONFIG_LIBDIR`](https://github.com/Homebrew/brew/blob/2.1.15/Library/Homebrew/extend/ENV/std.rb#L23)) is still broken, which is a regression from a few weeks ago. I'm still trying to figure out what changed in homebrew that they haven't reverted

scpeters:
I just reverted the patch because it wasn't helping and disabled pkg-config tests on master in a65e252568c3940c6f0fd3b9ecd526e3897b9860

scpeters:
[Build Status](https://build.osrfoundation.org/job/generic-release-homebrew_bottle_builder/754/) https://build.osrfoundation.org/job/generic-release-homebrew_bottle_builder/754/

Bo98:
> (which uses homebrew's Stdenv and Stdenv explicitly overrides PKG_CONFIG_LIBDIR)

Ah I forgot stdenv sets `PKG_CONFIG_LIBDIR`.

It's a bit of a superenv-type thing to do, but I won't question why it does that over `PKG_CONFIG_PATH`.

Homebrew/brew#7323 should fix this. I should have reverted that at the same time as the other things - sorry for the trouble.

scpeters:
thanks @Bo98 !

I'll try re-enabling the pkg-config tests after that gets merged

scpeters:
forgot to disable the test for gazebo11, so did that in e7390955528e23d51365a075d022392195aac377 and will rebuild bottles

scpeters:
[Build Status](https://build.osrfoundation.org/job/generic-release-homebrew_bottle_builder/755/) https://build.osrfoundation.org/job/generic-release-homebrew_bottle_builder/755/

follower:
> It's a bit of a superenv-type thing to do, but I won't question why it does that over `PKG_CONFIG_PATH`.

@Bo98, FWIW, my understanding of the reason for setting `PKG_CONFIG_LIBDIR` is that it prevents the system default search paths from being used (via `man pkg-config`):

```
       PKG_CONFIG_LIBDIR
              Replaces  the  default   pkg-config   search   directory,   usually
              /usr/lib/pkgconfig:/usr/share/pkgconfig.
```

Because "_The default directory will always be searched..._":

```
       PKG_CONFIG_PATH
              A  colon-separated (on Windows, semicolon-separated) list of direc-
              tories to search for .pc files.  The default directory will  always
              be   searched   after   searching   the   path;   the   default  is
              libdir/pkgconfig:datadir/pkgconfig where libdir is the  libdir  for
              pkg-config  and  datadir  is the datadir for pkg-config when it was
              installed.
```

Until I read [this](https://github.com/Homebrew/brew/blob/2bf8015bc44878d432fb9da4f11e1d8abcef8f32/Library/Homebrew/extend/ENV/std.rb#L21-L23), I misunderstood & thought setting `PKG_CONFIG_LIBDIR` overrode `PKG_CONFIG_PATH` too but that's not the case:

```ruby
    # Set the default pkg-config search path, overriding the built-in paths
    # Anything in PKG_CONFIG_PATH is searched before paths in this variable
    self["PKG_CONFIG_LIBDIR"] = determine_pkg_config_libdir
```

----

Drive-by comment after encountering this PR while researching https://github.com/Homebrew/homebrew-core/issues/56402#issuecomment-646304592. :)

PR #1002, opened 2020-04-19
ogre1.9: depends on macOS mojave

scpeters:


scpeters:
this is needed for #1001

j-rivero:
I've tried to reproduce the problem with opengl by removing the restriction on highsierra. However seems like we need to brew the qt4 tab (I did that in our CI code) but when [running test-bot it is uninstalling it](https://build.osrfoundation.org/job/generic-release-homebrew_bottle_builder/label=osx_highsierra/808/console).

scpeters:
I fixed the issue with the qt4 tap in https://github.com/scpeters/homebrew-test-bot/commit/a1dd15ccf45e54c34afa9a7c8f7dabd2f27f8d3c of our `still_working` branch and upstream at https://github.com/Homebrew/homebrew-test-bot/pull/426

PR #1004, opened 2020-05-08
ignition-physics2 2.1.0

osrfbuild:

PR #1010, opened 2020-05-14
ignition-msgs5 5.2.0

osrfbuild:


chapulina:
I see this had the same error as #1004 and I see @scpeters manually triggering bottle builds in https://build.osrfoundation.org/job/generic-release-homebrew_bottle_builder/.

Let me know if I can help with anything.

scpeters:
there was  a networking problem with the omnidroid build machine, and in the past I've had to unplug and replug the networking cable; so I did that this evening, and it seemed to fix things. then I restarted the bottle build

it's always something it seems

PR #1027, opened 2020-05-22
Remove broken bottles for: gui2, msgs1, sensors2, transport4, transport7. gazebo9 and gazebo10

j-rivero:
Protobuf bump killed the ABI in some cases, Bullet bump in some others. They should be regenerated, removing them to use brew test-bot without failures while testing code.

### DIFF
--- a/Formula/gazebo10.rb
+++ b/Formula/gazebo10.rb
@@ -7,12 +7,6 @@ class Gazebo10 < Formula
 
   head "https://bitbucket.org/osrf/gazebo", :branch => "default", :using => :hg
 
-  bottle do
-    root_url "https://osrf-distributions.s3.amazonaws.com/bottles-simulation"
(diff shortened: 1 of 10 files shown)


scpeters:
it looks like the `msgs1` dependency chain is properly cleared by this list, but I see some users of sensors2 and gui2 that still have bottle blocks:

~~~
$ grep gui2 $(grep -rlI 'bottle do' Formula/ | sort)
Formula//ignition-blueprint.rb:  depends_on "ignition-gui2"
Formula//ignition-gazebo2.rb:  depends_on "ignition-gui2"
Formula//ignition-launch1.rb:  depends_on "ignition-gui2"
~~~

I would include those 3 formulae as well

j-rivero:
> it looks like the `msgs1` dependency chain is properly cleared by this list, but I see some users of sensors2 and gui2 that still have bottle blocks:

I had the feeling that I miss some of the packages. Thanks!

j-rivero:
lets the jazz begin

scpeters:
👍

PR #1036, opened 2020-06-01
Style and format fixes to make test-bot happy

j-rivero:
The first set of changes won't fix all issues. Seems like brew is removing the dependencies on custom build flags (see https://github.com/Homebrew/brew/pull/2482), not sure how we want to react to that change. @scpeters ?

I think that the rest of changes are somehow trivial and should not affect how Formulas work.


scpeters:
it's not quite happy yet, but this is a good start, thanks!

PR #1037, opened 2020-06-02
Bitbucket -> GitHub in head, home fields

scpeters:
Also updates the branch names when relevant.

I think CI is broken, so I expect this to be red.

### DIFF
--- a/Formula/gazebo3.rb
+++ b/Formula/gazebo3.rb
@@ -3,7 +3,7 @@ class Gazebo3 < Formula
   homepage "http://gazebosim.org"
   url "https://osrf-distributions.s3.amazonaws.com/gazebo/releases/gazebo-3.2.0.tar.bz2"
   sha256 "6b77382a491833d5292b3e1fca34a04c968025a09746d87cdcf77cff040acea5"
-  head "https://github.com/osrf/gazebo", :branch => "master"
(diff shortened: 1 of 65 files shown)


Formula/gazebo3.rb
```diff
@@ -3,7 +3,7 @@ class Gazebo3 < Formula
   homepage "http://gazebosim.org"
   url "https://osrf-distributions.s3.amazonaws.com/gazebo/releases/gazebo-3.2.0.tar.bz2"
   sha256 "6b77382a491833d5292b3e1fca34a04c968025a09746d87cdcf77cff040acea5"
-  head "https://bitbucket.org/osrf/gazebo", :branch => "default", :using => :hg
+  head "https://github.com/osrf/gazebo", :branch => "master"
```
j-rivero, line 6: we don't have a `gazebo3` branch but `master` is not probably the code what a Gazebo archeologist would be looking for.

scpeters:
I am planning to deprecate many of these old formulae, since brew recently added a `:deprecated` tag in https://github.com/Homebrew/brew/pull/7190

Formula/gazebo3.rb
scpeters: thanks, I was being lazy but I fixed it in 6c43a6bba443c8a5967f548ed173fe17de55d8c8

PR #1054, opened 2020-06-11
sdformat10: new formula, depends on tinyxml2

scpeters:
Builds from head of `sdf10` branch https://github.com/osrf/sdformat/commit/1e213886f892e6accd1c7b0f104a8af2583ed68b

### DIFF
--- a/Formula/sdformat10.rb
+++ b/Formula/sdformat10.rb
@@ -1,7 +1,6 @@
 class Sdformat10 < Formula
   desc "Simulation Description Format"
   homepage "http://sdformat.org"
-  url "https://osrf-distributions.s3.amazonaws.com/sdformat/releases/sdformat-9.2.0.tar.bz2"
   url "https://github.com/osrf/sdformat/archive/1e213886f892e6accd1c7b0f104a8af2583ed68b.tar.gz"
   version "9.999.999~0~20200604~1e21388"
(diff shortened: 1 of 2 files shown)


Formula/sdformat10.rb
```diff
@@ -0,0 +1,70 @@
...
+  version "9.999.999~0~20200604~1e21388"
+  sha256 "8b33c4675c7447a780d676bf40c50925524d210982f857d6ffb17a2279f745ca"
+
+  depends_on "cmake" => [:build, :test]
+  depends_on "pkg-config" => [:build, :test]
+
+  depends_on "doxygen"
+  depends_on "ignition-math6"
+  depends_on :macos => :mojave # c++17
+  depends_on "tinyxml"
```
j-rivero, line 14: are we using both tinyxml for sdformat? or this one should be removed?

scpeters: similar to my comment in https://github.com/ignition-tooling/release-tools/pull/220#discussion_r439417519, it keeps everything working in CI if both tinyxml and tinyxml2 are installed until https://github.com/osrf/sdformat/pull/264 is merged. this formula is based on the head of the `sdf10` branch, so it actually needs `tinyxml` and not `tinyxml2`, but adding them both allows CI to work for either set of branches

I'll add a comment that the `tinyxml` dependency should be removed after https://github.com/osrf/sdformat/pull/264 is merged

scpeters: dd52ace

PR #1056, opened 2020-06-13
Deprecate old gazebo formulae, fix style

scpeters:


scpeters:
wow, CI is green again!

PR #1065, opened 2020-07-16
ignition-transport9: new formula

chapulina:
Remove alias and add new formula that depends on `ign-msgs6`. I took the last commit of https://github.com/ignitionrobotics/ign-transport/pull/149/

This is broken off https://github.com/osrf/homebrew-simulation/pull/1063 so we can make the changes incrementally.

PR #1070, opened 2020-07-21
ignition-transport9: fix version

chapulina:
So, I messed up on #1065 and `ignition-transport9` was released with a major too high. Now brew won't allow a version decrease.

Is this fixable? 😟


scpeters:
yes, this is fixable if we add `version_scheme 1`

* https://github.com/osrf/homebrew-simulation/blob/master/Formula/ignition-msgs1.rb#L8

PR #1082, opened 2020-07-30
Remove gazebo[1-5] and sdformat[1-2] Formulas/Aliases

chapulina:
All the latest PRs are failing at `brew test-bot` with lots of "Use the new Ruby 1.9 hash syntax" warnings.

So I used this tip to convert all hashes: 

http://effectif.com/ruby/update-your-project-for-ruby-19-hash-syntax


chapulina:
Now it's stopping at `chrono-engine`

j-rivero:
A ghost error appeared now:

```
Error: Failure while executing; `brew deps --full-name --include-build --include-test osrf/simulation/gazebo2` exited with 1. Here's the output:
##[error]Process completed with exit code 1.
```
I'm trying to add qt4 tap to see if that is the reason.

j-rivero:
ok, it's failing with problems on cmake code for sdformat 2.3.2. I think is a good moment to reduce the testing times and the maintenance overload: 8701e1c removes sdformat 2.3.2 and gazebo versions from 2 to 5 (already end of life).

j-rivero:
```
Error: 1 problem in 1 formula detected
osrf/simulation/ignition-fuel-tools3:
  * 'revision 1' should be removed
```
Ugh, this is totally new to me and not sure why is an error. @scpeters any idea of this one?

scpeters:
we can fix these syntax errors with `brew style --fix`. I'll submit a separate PR based on that and then see which errors are still occurring

j-rivero:
> Ugh, this is totally new to me and not sure why is an error. @scpeters any idea of this one?

the problems seems to persists. Should the `revision 1` tag be removed. I can not find any documentation about this issue.

scpeters:
> > Ugh, this is totally new to me and not sure why is an error. @scpeters any idea of this one?
> 
> the problems seems to persists. Should the `revision 1` tag be removed. I can not find any documentation about this issue.

this looks like a bug in brew to me in the following code:

https://github.com/Homebrew/brew/blob/2.4.9/Library/Homebrew/dev-cmd/audit.rb#L821-L825

scpeters:
I think it's fine to delete the old gazebo formulae that have `deprecate!` statements (from #1056), but I'd like to add one of these statements to sdformat[1-2] and wait a month or so before deleting them

chapulina:
> I'd like to add one of these statements to sdformat[1-2] and wait a month or so before deleting them

That sounds reasonable. I'd also suggest deprecating Ignition versions used by deprecated Gazebo and Acropolis. I'll send you a list, @scpeters .

scpeters:
> > I'd like to add one of these statements to sdformat[1-2] and wait a month or so before deleting them
> 
> That sounds reasonable. I'd also suggest deprecating Ignition versions used by deprecated Gazebo and Acropolis. I'll send you a list, @scpeters .

so there are methods for `deprecate!` and `disable!` that take a date argument. maybe we should use the `disable!` keyword to indicate when we plan to delete these, to give users notice. I'll take care of this when you send a list (or make an issue with the list)

chapulina:
> I'll take care of this when you send a list (or make an issue with the list)

I shared the spreadsheet with you and ticketed this issue about making that info public: https://github.com/ignitionrobotics/docs/issues/49

chapulina:
I'm not sure about the status of this PR, so I'm closing it. At this point I think it would be cleaner to open a new one to deprecate old versions.

PR #1086, opened 2020-08-03
Update msgs6 to latest commit

j-rivero:
Need latest msgs6 features for https://github.com/ignitionrobotics/ign-gazebo/pull/262


j-rivero:
Problems in CI are being worked out in #1082

PR #1124, opened 2020-09-14
ignition-math6: revert swig dependency

scpeters:
It's causing CI failures, so revert it for now.


scpeters:
needed by https://github.com/ignitionrobotics/ign-math/pull/158

chapulina (approved):

PR #1173, opened 2020-10-05
New formula: Ignition Edifice

chapulina:
At the moment, the dependencies are the same as Dome.

Here's the upstream repo: https://github.com/ignitionrobotics/ign-edifice


scpeters:
looks good, and this doesn't need a bottle

PR #1183, opened 2020-10-14
ignition-rendering3 3.2.0

osrfbuild:


chapulina:
build bottle

scpeters:
I did the same thing earlier today, but I don't think we need new bottles because this is a re-run of `release.py` that has uploaded a new tarball (see the diff)

the proper thing to do here is just merge the PR without a new bottle

scpeters:
I just cancelled the job

chapulina:
Ok, I see. So a new hash doesn't prompt a new bottle :ok_hand:

scpeters:
> Ok, I see. So a new shasum doesn't prompt a new bottle 👌

unless the tag is actually changed, but in this case I believe the tag is still pointing to the same commit, so the contents of the tarball are nominally the same, but differ due to timestamps

PR #1203, opened 2020-11-16
Patch ignition-cmake2 and rebuild ignition-rendering4 bottle

scpeters:
This patches ignition-cmake2 with https://github.com/ignitionrobotics/ign-cmake/pull/125 to fix `FindIgnOGRE2`, which should fix https://github.com/ignitionrobotics/ign-sensors/issues/62 after we have rebuilt the corresponding ignition-rendering bottles. This PR will rebuild `ignition-rendering4`. If it is successful, I will rebuild the other supported versions as well.


scpeters:
build bottle

PR #1219, opened 2020-12-07
Add tests for ignition collections

scpeters:
Use `vcs validate` to test the gazebodistro yaml files.

PR #1220, opened 2020-12-07
ignition-blueprint: test with vcs validate, update commit

scpeters:
This adds a test of the `ignition-blueprint` yaml file similar to #1219, but it also updates the formula to point to the latest commit, since the current release still points to bitbucket:

* https://github.com/ignitionrobotics/ign-blueprint/blob/ignition-blueprint_1.0.0/gazebodistro/collection-blueprint.yaml


scpeters:
build bottle

PR #1221, opened 2020-12-08
ignition-gazebo*: test `ign gazebo -s`

scpeters:
This requires dependencies on ruby and ignition-tools.


scpeters:
build bottle

PR #1237, opened 2020-12-21
ignition-utils0: new formula

chapulina:
Used the commit from https://github.com/ignitionrobotics/ign-utils/pull/1


mjcarroll (approved):

Formula/ignition-utils0.rb
```diff
@@ -0,0 +1,53 @@
...
+    system "cmake", ".", *std_cmake_args
+    system "make", "install"
+  end
+
+  test do
+    (testpath/"test.cpp").write <<-EOS
+      #include <ignition/utils/ImplPtr.hh>
+      class SomeClass
+      {
+        private: ignition::utils::ImplPtr<Implementation> dataPtr;
```
mjcarroll, line 23: You'll need an `Implementation` somewhere above for this to work.

chapulina: fd7c0bd0ed4e37a6c95ac2aaf2f26306e5b3a853

scpeters:
build bottle

scpeters:
build bottle

scpeters:
test is failing

~~~
test.cpp:10:19: error: call to implicitly-deleted default constructor of 'SomeClass'
        SomeClass object;
                  ^
test.cpp:7:61: note: default constructor of 'SomeClass' is implicitly deleted because field 'dataPtr' has no default constructor
        private: ignition::utils::ImplPtr<SomeClassPrivate> dataPtr;
                                                            ^
1 error generated.
~~~

cc @mjcarroll

Formula/ignition-utils0.rb
```diff
@@ -0,0 +1,60 @@
...
+    system "make", "install"
+  end
+
+  test do
+    (testpath/"test.cpp").write <<-EOS
+      #include <ignition/utils/ImplPtr.hh>
+      class SomeClassPrivate
+      {
+      };
+      class SomeClass
```
azeey, line 26: I think `dataPtr` has to be constructed using `MakeImpl`
```suggestion
      class SomeClass: dataPtr(ignition::utils::MakeImpl<SomeClassPrivate>())
```

scpeters:
build bottle

Formula/ignition-utils0.rb
scpeters: that's the class definition, not the constructor

* [Build Status](https://build.osrfoundation.org/job/generic-release-homebrew_triggered_bottle_builder/label=osx_mojave/189/) https://build.osrfoundation.org/job/generic-release-homebrew_triggered_bottle_builder/label=osx_mojave/189/console

azeey: Oops, sorry, it should have been
```c++
class SomeClass
{
  SomeClass(): dataPtr(ignition::utils::MakeImpl<SomeClassPrivate>()) 
  {}
...
```

scpeters: trying member initializer in d552c9b

scpeters:
build bottle

Formula/ignition-utils0.rb
scpeters: > Oops, sorry, it should have been

sorry I pushed before I saw your comment. Do you think it's better to define the default constructor or use an inline member initializer?

scpeters:
now the first compilation that calls clang directly is working, but it fails when building with cmake. After noticing that we are passing `-std=c++17` to clang, I did some debugging and it looks like our cmake config isn't exporting the c++17 requirement for ignition-utils0. I'll post on the pull request about this

Formula/ignition-utils0.rb
azeey: The most common use case is going to be via the default constructor as shown in https://github.com/osrf/sdformat/pull/447/files#diff-a6e2168710bc843df1acd00632454dbe90626aa4f0e21cef75c9c0027d15a482R42, but I don't think it matters much for this test.

scpeters: :+1:

scpeters:
build bottle

PR #1254, opened 2020-12-28
ignition-gazebo2 2.26.0

osrfbuild:

PR #1258, opened 2020-12-30
Make doxygen an optional build dependency

chapulina:
This is an attempt at fixing https://github.com/ignitionrobotics/ign-gazebo/issues/520 and unblocking https://github.com/osrf/homebrew-simulation/pull/1254

I think these are the Blueprint / Citadel / Dome formulae which may be bringing doxygen to `ign-gazebo` builds.

### DIFF
--- a/Formula/ignition-math6.rb
+++ b/Formula/ignition-math6.rb
@@ -13,7 +13,7 @@ class IgnitionMath6 < Formula
   end
 
   depends_on "cmake" => :build
-  depends_on "doxygen" => :build
+  depends_on "doxygen" => [:build, :optional]
   depends_on "eigen"
   depends_on "ignition-cmake2"
   depends_on "ruby"
--- a/Formula/ogre1.9.rb
+++ b/Formula/ogre1.9.rb
@@ -15,8 +15,8 @@ class Ogre19 < Formula
   option "with-cg"
 
   depends_on "cmake" => :build
-  depends_on "boost"
   depends_on "doxygen" => [:build, :optional]
+  depends_on "boost"
   depends_on "freeimage"
   depends_on "freetype"
   depends_on "libx11"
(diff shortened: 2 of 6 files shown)


scpeters:
I made an alternative suggestion in https://github.com/ignitionrobotics/ign-gazebo/issues/520

scpeters:
optional dependencies are now discouraged by brew, so I wouldn't recommend adding them here

I think we can be more consistent in how we treat `doxygen` dependencies, but here's my understanding:

1. a `:build` depend on `doxygen` will make it available when building homebrew bottles, but not install it with `brew install --only-dependencies`, which we use for CI. So for example, doxygen is [not found by the ign-math6 job](https://build.osrfoundation.org/view/os_homebrew/job/ignition_math-ci-ign-math6-homebrew-amd64/63/console)

2. a regular depend on `doxygen` will make it available for our CI jobs to the named package and all its dependents.

I think the proper thing to do would be to always declare `doxygen` as a `:build` dependency in formulae, and also add it to `BREW_BASE_DEPENDENCIES` in [dependencies_archive.sh](https://github.com/ignition-tooling/release-tools/blob/07aa0f70332ebe4005a684eaa1f7f84827d1f853/jenkins-scripts/lib/dependencies_archive.sh#L53) to it will be tested in CI (if desired). we could also add logic to `brew unlink doxygen` for `ignition-gazebo` CI jobs on homebrew

chapulina:
> optional dependencies are now discouraged by brew

Oh ok, I mainly used it because I saw in other formulae, I think it's fine without it

> a :build depend on doxygen will make it available when building homebrew bottles, but not install it with brew install --only-dependencies, which we use for CI

Argh I see, it sounds like brew could use some `:test` dependencies. I didn't mean for this to become too complicated, I'll manage this on `ign-gazebo` instead then.

PR #1278, opened 2021-01-11
ignition-edifice: bump dependencies

chapulina:
Addressing https://github.com/ignition-tooling/release-tools/issues/362

Uses commit from https://github.com/ignitionrobotics/ign-edifice/pull/2/


nkoenig (approved):

scpeters:
build bottle

scpeters:
build bottle

chapulina:
I don't know why it would think that the previous version was 8:

```
osrf/simulation/ignition-edifice:
  * stable version should not decrease (from 8 to 0.999.999~0~20210111~be4570)
```

scpeters:
> I don't know why it would think that the previous version was 8:
> 
> ```
> osrf/simulation/ignition-edifice:
>   * stable version should not decrease (from 8 to 0.999.999~0~20210111~be4570)
> ```

it was originally merged without a `version` attribute, so it inferred the version from the tarball url, probably some random number from the commit hash. the solution is to bump the `version_scheme` (see dd54e3f1e8fbadd1c239921cd1593bf96c717c54 for example)

chapulina:
build bottle

PR #1281, opened 2021-01-12
gazebo11: revision bump for bullet

scpeters:
The most recent homebrew bottle for bullet switched to double-precision, but this was reverted in https://github.com/Homebrew/homebrew-core/pull/68634, so we need to rebuild bottles to fix the broken ABI.


scpeters:
build bottle

scpeters:
build bottle

PR #1284, opened 2021-01-12
Bump in edifice: ign-gazebo5

chapulina:

Uses commit from https://github.com/ignitionrobotics/ign-gazebo/pull/549


scpeters:
this will be handled by #1287

PR #1285, opened 2021-01-12
ignition-launch4: depend on sdformat11 and ignition-physics4

chapulina:


scpeters:
this will be handled by #1287

PR #1286, opened 2021-01-12
ignition-physics4: new formula

chapulina:

### DIFF
--- a/Aliases/ignition-physics4
+++ b/Aliases/ignition-physics4
@@ -1,1 +1,0 @@
-../Formula/ignition-physics3.rb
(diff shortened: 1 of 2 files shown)


scpeters:
build bottle

scpeters:
I like putting new formulae in their own PR, so this is good, but after this I think we'll need to combine the other PR's that change dependencies of formulae in the same chain (such as #1284, #1285, #1287 and #1288). I made it work for msgs7/transport10 by deleting the bottles and then rebuilding, but I think it will be easier if we combine those branches and rebuild the affected bottles all at once

PR #1311, opened 2021-01-28
sdformat11: depend on ignition-utils1

azeey:


chapulina:
I'm not sure if it's best to update the url and shasum to point to the last commit on https://github.com/osrf/sdformat/pull/474. I always assumed so, but now looking for an example I saw that @scpeters didn't do that on https://github.com/osrf/homebrew-simulation/pull/1213/. I'm also surprised that no bottle was build on that PR, my intuition would be to at least bump the revision here and build a new bottle.

scpeters:
> I'm not sure if it's best to update the url and shasum to point to the last commit on [osrf/sdformat#474](https://github.com/osrf/sdformat/pull/474). I always assumed so, but now looking for an example I saw that @scpeters didn't do that on #1213. I'm also surprised that no bottle was build on that PR, my intuition would be to at least bump the revision here and build a new bottle.

sometimes there's a pressing need to build a bottle, like if you are changing a dependency from one version to another. but if you are adding a dependency, it's more flexible. simply adding the dependency to the formula allows CI to pass, without having to update the commit. I don't think this has a large effect on any downstream packages, so the one-line change seems fine to me

PR #1316, opened 2021-02-01
Use new bottle syntax

scpeters:
I used `brew style --fix osrf/simulaton` with the branch from https://github.com/Homebrew/brew/pull/10453 that is still under review.


scpeters:
seems to work fine on my machine

PR #1458, opened 2021-05-13
README: add edifice bottle badges

scpeters:


chapulina:
Missing `ign-utils` and `libsdformat` :wink:

scpeters:
> Missing `ign-utils` and `libsdformat`

so there is an ign-utils `install_bottle` job, but it's not part of the edifice view. I'll add its badge here

we don't have any `install_bottle` jobs for sdformat

scpeters:
> > Missing `ign-utils` and `libsdformat`
> 
> so there is an ign-utils `install_bottle` job, but it's not part of the edifice view. I'll add its badge here
> 
> we don't have any `install_bottle` jobs for sdformat

https://github.com/ignition-tooling/release-tools/pull/460

scpeters:
I've added badges for sdformat and ign-utils

chapulina (approved):

Issue #1463, opened 2021-05-18
Rebuild bottles for protobuf 3.17.0


scpeters:
The protobuf 3.17.0 bottle is pending in https://github.com/Homebrew/homebrew-core/pull/77373 and will break all our bottles once it is merged. We should do the following:

1. Remove bottle blocks for everything downstream of protobuf, staring from `ignition-msgs*` and `gazebo*` so that they will build from source until bottle are rebuilt
2. Rebuild bottles incrementally.
 - [x] msgs
 - [x] fuel-tools
 - [x] transport
 - [x] gui
 - [x] sensors
 - [x] gazebo
 - [x] launch
 - [x] collections
 - [x] gazebo9
 - [x] gazebo11

j-rivero:
I think I can take this one, the homebrew-core was merged some minutes ago.

> Rebuild bottles incrementally.

Does it makes sense to use the same pull request to do it? Changing several revisions at once will work?

scpeters:
> I think I can take this one, the homebrew-core was merged some minutes ago.
> 
> > Rebuild bottles incrementally.
> 
> Does it makes sense to use the same pull request to do it? Changing several revisions at once will work?

it can work, but an sporadic infrastructure problem is more likely to affect them all and make the builds longer. it's a bit safer to cluster them together: one for each ignition stack (citadel, dome, edifice, fortress), and one (or 2) for gazebo9 / 11. it is more effort, but in my experience with build failures, it's nicer to segregate the packages

scpeters:
since all the bottles are gone at the moment, another alternative is to increment all the ign-msgs* versions in a single PR, then ign-transport* and go gradually that way. we could try to rebuild them all at once and see if it works, but it may take 12+ hours

scpeters:
thanks for the help @j-rivero; I'll let you take it from here

scpeters:
the build farm is clear so I got it started with #1466

PR #1462, opened 2021-05-18
ignition-rendering6: build main branch from source

scpeters:
Similar to #1460, part of #1314, needed since https://github.com/ignitionrobotics/ign-rendering/pull/314 was merged.

PR #1460, opened 2021-05-14
ignition-physics5: build main branch from source

chapulina:
Similar approach to https://github.com/osrf/homebrew-simulation/pull/1459

Needed by https://github.com/ignitionrobotics/ign-gazebo/pull/811

PR #1459, opened 2021-05-14
sdformat12: build main branch from source

scpeters:
This is a potential approach for solving #1314, motivated by https://github.com/ignitionrobotics/ign-gazebo/pull/811, which has failing CI since the sdformat12 formula is outdated.

PR #1469, opened 2021-05-19
Bump ign-transport* for protobuf 3.17.0

j-rivero:
Part of #1463


j-rivero:
build bottle

scpeters:
I skipped `ignition-msgs8` in #1466 so we could change it to point to the `main` branch instead of a specific commit (see #1314, #1459, #1460, #1462). I would recommend handling the fortress dependencies separately, so we could start by removing `ignition-transport11` from this PR

j-rivero:
build bottle

Issue #1477, opened 2021-05-21
gazebo bottles broken due to new tbb version


mcjsk:
I installed `gazebo11` with the default [one-liner][0] on my MBP when running Catalina, and Gazebo worked.

I just upgraded to Big Sur, then ran:

`brew update && brew upgrade`

Now `gazebo` fails to run with:

```sh
~ » gazebo
dyld: Symbol not found: __ZNSt12length_errorD1Ev
  Referenced from: /usr/local/lib/libgazebo_common.11.dylib
  Expected in: /usr/local/opt/tbb/lib/libtbb.dylib
 in /usr/local/lib/libgazebo_common.11.dylib
zsh: abort      gazebo
```

`brew doctor` output:

```sh
~ » brew doctor
Please note that these warnings are just used to help the Homebrew maintainers
with debugging if you file an issue. If everything you use Homebrew for is
working fine: please don't worry or file an issue; just ignore this. Thanks!

Warning: You have the following deprecated, cask taps tapped:
  caskroom/cask
Untap them with `brew untap`.

Warning: "config" scripts exist outside your system or Homebrew directories.
`./configure` scripts often look for *-config scripts to determine if
software packages are installed, and which additional flags to use when
compiling and linking.

Having additional scripts in your path can confuse software installed via
Homebrew if the config script overrides a system or Homebrew-provided
script of the same name. We found the following "config" scripts:
  /Library/Frameworks/Python.framework/Versions/3.7/bin/python3.7-config
  /Library/Frameworks/Python.framework/Versions/3.7/bin/python3.7m-config
  /Library/Frameworks/Python.framework/Versions/3.7/bin/python3-config

Warning: You have unlinked kegs in your Cellar.
Leaving kegs unlinked can lead to build-trouble and cause formulae that depend on
those kegs to fail to run properly once built. Run `brew link` on these:
  python@3.9

Warning: Homebrew's "sbin" was not found in your PATH but you have installed
formulae that put executables in /usr/local/sbin.
Consider setting your PATH for example like so:
  echo 'export PATH="/usr/local/sbin:$PATH"' >> ~/.zshrc
```

`brew config` output:

```sh
~ » brew config
HOMEBREW_VERSION: 3.1.8
ORIGIN: https://github.com/Homebrew/brew
HEAD: e0bfcdaf258dadaaf8c102d4cb51f23b6fdb2c4c
Last commit: 3 days ago
[...]

PR #1475, opened 2021-05-21
ignition-launch: bump revision for protobuf

j-rivero:
Part of #1463

PR #1478, opened 2021-05-21
gazebo*: remove bottles due to tbb

scpeters:
Part of #1477.


scpeters:
the mojave build has already started for #1475, so I believe this is safe to merge

PR #1479, opened 2021-05-21
gazebo11: use tbb@2020 and update test

scpeters:
I've changed the test to include `gazebo/gazebo.hh`, which includes tbb headers via [gazebo/transport/Connection.hh](https://github.com/osrf/gazebo/blob/gazebo11_11.5.1/gazebo/transport/Connection.hh#L57). Sadly, we aren't exporting the `tbb` dependency in our cmake config, so we need to use `CPATH` to properly include the tbb header files.


scpeters:
try building this after #1475 completes

scpeters:
build bottle

PR #1490, opened 2021-06-09
Disable bottles that depend on tinyxml2

mjcarroll:
Signed-off-by: Michael Carroll <michael@openrobotics.org>

PR #1493, opened 2021-06-09
tinyxml2 fallout: unbottle dependents of msgs5, msgs6, msgs7

j-rivero:
Follow up to #1490


scpeters:
I think we missed removing the bottle for ignition-transport8

PR #1487, opened 2021-06-07
sdformat6 6.3.0

osrfbuild:

PR #1495, opened 2021-06-09
Rebuild sdformat 6.3.0 bottles (SHA256 hash mismatch)

j-rivero:
I think that Jenkins died before finishing successfully #1487 . Let's rebuild the bottles:


j-rivero:
build bottle

PR #1501, opened 2021-06-10
tinyxml2 fallout: rebuild gazebo9 bottle

j-rivero:
Follow up to #1490


j-rivero:
There is a new version of Gazebo9 coming today or tomorrow. I think we can cancel this one.

PR #1513, opened 2021-06-17
[dart] Bump commit

chapulina:
I don't know if this is the correct thing to do. I'm trying to sync the macOS binaries with the latest Ubuntu ones to fix https://github.com/ignitionrobotics/ign-physics/pull/262/.

### DIFF
--- a/Formula/dartsim@6.10.0.rb
+++ b/Formula/dartsim@6.10.0.rb
@@ -1,12 +1,12 @@
 class DartsimAT6100 < Formula
   desc "Dynamic Animation and Robotics Toolkit (openrobotics port)"
   homepage "https://dartsim.github.io/"
-  # osrc custom nightly built from commit fdde7e7894ebc36bae8811f7a63e5b1c899bb4af
-  url "https://github.com/azeey/dart/archive/1673b0be51fb370023df7490dc49706b590d8f72.tar.gz"
-  version "6.10.0~20200916~1673b0be51fb370023df7490dc49706b590d8f72"
-  sha256 "ec2e833d3225ac3f4365cc6d8b2f5511170a47d140ff43dcc7d63a50fbb6bfd5"
+  # OSRF's fork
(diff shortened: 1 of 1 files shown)


chapulina:
build bottle

Formula/dartsim@6.10.0.rb
```diff
@@ -1,12 +1,12 @@
 class DartsimAT6100 < Formula
   desc "Dynamic Animation and Robotics Toolkit (openrobotics port)"
   homepage "https://dartsim.github.io/"
-  # osrc custom nightly built from commit fdde7e7894ebc36bae8811f7a63e5b1c899bb4af
-  url "https://github.com/azeey/dart/archive/1673b0be51fb370023df7490dc49706b590d8f72.tar.gz"
-  version "6.10.0~20200916~1673b0be51fb370023df7490dc49706b590d8f72"
-  sha256 "ec2e833d3225ac3f4365cc6d8b2f5511170a47d140ff43dcc7d63a50fbb6bfd5"
+  # OSRF's fork
+  url "https://github.com/ignition-forks/dart/archive/71c7cfb97e8b583aa1ada639b6d1cf9a1835c1da.tar.gz"
```
scpeters, line 5: I believe the debs are built from https://github.com/ignition-forks/dart/commit/175596b511adb55d831c0c3401ef50e2cb96a6f3

scpeters: I'll update this

scpeters: f151eb2

scpeters:
build bottle

Formula/dartsim@6.10.0.rb
chapulina: Thanks. The extra commit was just adding the changelog, so I think the end result won't be different.

scpeters: agreed, but this way it's easier to match up the debian and homebrew versions

PR #1526, opened 2021-06-30
ignition-gazebo5 5.1.0~pre1

osrfbuild:

PR #1528, opened 2021-06-30
ignition-gazebo5 5.1.0

osrfbuild:


scpeters:
there was an error when trying to build #1526 since ignition-sensors5 doesn't have a bottle; so let's add a revision bump for that as well to trigger a new bottle

scpeters:
build bottle

Issue #1314, opened 2021-01-30
Nightlies for main branches on homebrew


chapulina:
Our debian nightlies are very convenient to get fresh debians from the `main` branches of all Ignition libraries. It would be great to have the same for homebrew. 

At the moment, it's still necessary to open a pull request here bumping the commit version of main branches when they're needed by downstream libraries. This could be automated, with GitHub actions for example, to happen nightly.

scpeters:
We have separate repositories in packages.osrfoundation.org for stable, prerelease, and nightly packages, but for homebrew we only have one version of things, which typically points to the latest stable release or prerelease. Before a formula has had any releases, we just point to the tarball for a specific commit on github and update it manually as needed.

I assume this proposal would apply only to packages that have not yet had a prerelease or stable release?

chapulina:
> I assume this proposal would apply only to packages that have not yet had a prerelease or stable release?

Correct, just like the nightlies, this would only apply to the `main` branches. Although in some cases there will be overlap between pre-release and nightly, especially as we start shipping pre-releases close to the stable release date.

chapulina:
> just like the nightlies, this would only apply to the main branches.

The debian nightlies are now using Fortress branches instead of `main`, see https://github.com/ignition-tooling/release-tools/pull/432

> I assume this proposal would apply only to packages that have not yet had a prerelease or stable release?

The homebrew nightlies would either:

* Use `main` branches, which are considered unstable and haven't had stable releases.
    * This should leave no ambiguity about the status of those packages.
    * We'd need to make pre- / stable releases  whenever we need a change from stable branches.
* Use Fortress branches, which consist of a combination of stable and unstable branches.
    * This would allow iterating faster on Fortress, like we're doing with debian nightlies.
    * This may introduce regressions on stable branches. We could consider this a good thing (detecting regressions early) or a bad thing (using homebrew as a canary).

scpeters:
One very easy option occurred to me for unreleased packages: use the git repository as the url instead of a tarball of an explicit commit. We already do this for `cppzmq` for example:

https://github.com/osrf/homebrew-simulation/blob/master/Formula/cppzmq.rb#L4-L5

This would actually be closer to our windows CI than Ubuntu nightlies, since it would build the `main` branch of unreleased packages from source. We may actually be able to build binaries, but I think it wouldn't be easy to know which commit was used since the `version` tag will not include the commit `sha`

scpeters:
[...]

PR #1529, opened 2021-06-30
sdformat12: build from source

chapulina:
Since https://github.com/osrf/homebrew-simulation/pull/1459 the bottle is being built from main, which is much easier than picking a commit and getting the hash.

But since a bottle is built, I think we still need to open this kind of PR to update the revision. Can we just not build a bottle, @scpeters ?

This is needed by https://github.com/ignitionrobotics/ign-gazebo/pull/888

### DIFF
--- a/Formula/sdformat12.rb
+++ b/Formula/sdformat12.rb
@@ -2,7 +2,7 @@ class Sdformat12 < Formula
   desc "Simulation Description Format"
   homepage "http://sdformat.org"
   url "https://github.com/osrf/sdformat.git", branch: "main"
-  version "11.999.999~2"
+  version "11.999.999~1"
   license "Apache-2.0"
   revision 1
 
--- a/Formula/sdformat12.rb
+++ b/Formula/sdformat12.rb
@@ -2,15 +2,9 @@ class Sdformat12 < Formula
   desc "Simulation Description Format"
   homepage "http://sdformat.org"
   url "https://github.com/osrf/sdformat.git", branch: "main"
-  version "11.999.999~1"
+  version "11.999.999~2"
   license "Apache-2.0"
   revision 1
-
-  bottle do
-    root_url "https://osrf-distributions.s3.amazonaws.com/bottles-simulation"
-    sha256 catalina: "3bb0ef4fc4774d980775dcb66a8fdf0eddca9dbfa62f63df4d7b591dcbbe0730"
-    sha256 mojave:   "630966a6f083fef0b9d59764cac893e530e77e54ad521625fa4b727c85b542e6"
-  end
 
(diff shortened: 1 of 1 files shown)


Formula/sdformat12.rb
```diff
@@ -2,16 +2,10 @@ class Sdformat12 < Formula
   desc "Simulation Description Format"
   homepage "http://sdformat.org"
   url "https://github.com/osrf/sdformat.git", branch: "main"
-  version "11.999.999~1"
+  version "11.999.999~2"
   license "Apache-2.0"
   revision 1
```
scpeters, line 7: I would leave the `version` as is and increment the `revision`

scpeters: you don't need to remove the bottle block either if we're going to rebuild them in this PR

chapulina: But is building the bottle a good idea? Could we just leave the unstable formulae being built from source, so we don't need to open PRs to update revisions?

scpeters: it's faster to use a bottle than to build from source, so it's a trade-off

chapulina: My concern is having to open these PRs, because they're not automated (yet) like our nightlies, because we're actively adding features to Fortress which are needed by downstream libraries.

I understood that was the resolution of #1314:

_@scpeters: This would actually be closer to our windows CI than Ubuntu nightlies, since it would build the main branch of unreleased packages from source._

scpeters: ok, I viewed it as a small amount of maintenance to rebuild bottles occasionally to keep fortress CI running faster, but I think we weren't on the same page. The following fortress formulae currently have bottles:

~~~
$ grep -l 'bottle do' $(grep -rlI 'url .* "main"' Formula/ | sort)
Formula//ignition-fuel-tools7.rb
Formula//ignition-msgs8.rb
Formula//ignition-physics5.rb
Formula//ignition-rendering6.rb
Formula//ignition-transport11.rb
Formula//sdformat12.rb
~~~

we don't need to change any version numbers or revisions; we just need to remove the bottle blocks for all these formulae and they will rebuild from source each time

chapulina: Cool, let's go with source builds to see if this allows us to iterate faster. If CI starts to take prohibitively long, we can revisit :+1:

chapulina:
Merging with just the bottle removal!

Formula/sdformat12.rb
scpeters: removing the other fortress bottles in https://github.com/osrf/homebrew-simulation/pull/1530

PR #1540, opened 2021-07-24
ignition-plugin2: new formula

scpeters:
Add dependency on ignition-utils1. Needed by https://github.com/ignitionrobotics/ign-plugin/pull/55


adlarkin (commented):
LGTM. I'll let someone else take a look at this as well before merging since I don't have too much experience with homebrew formulas

chapulina (approved):

PR #1558, opened 2021-08-31
ignition-gui[345]: rm qwt dependency

scpeters:
There is an open pull request to upgrade `qwt` to use qt version 6, so home-brew formulae that want to use `qwt` with `qt@5` should now depend on the `qwt-qt5` formula ( as of https://github.com/Homebrew/homebrew-core/pull/84275 ). These ignition-gui formulae list `qwt` as a dependency but don't actually use it, so just remove the dependency and bump revision so we can update the bottles.


scpeters:
build bottle

PR #1559, opened 2021-08-31
gazebo11: depend on qwt-qt5

scpeters:
There is an open pull request to upgrade `qwt` to use qt version 6, so home-brew formulae that want to use `qwt` with `qt@5` should now depend on the `qwt-qt5` formula ( as of https://github.com/Homebrew/homebrew-core/pull/84275 ). This updates `gazebo11` to use `qwt-qt5` and bumps the revision. I will do the same for `gazebo9` in a separate pull request.


scpeters:
build bottle

PR #1619, opened 2021-10-10
sdformat6: revision bump for urdfdom

scpeters:
Needed since urdfdom 3.0.0 was released in https://github.com/Homebrew/homebrew-core/pull/87001


scpeters:
build bottle

PR #1643, opened 2021-10-21
ogre1.9: update for compatibility with Apple M1 Chip

zarathustr:
The commit includes a full patch for successful compilation of ogre1.9 on the Apple M1 chip. THe main modifications are that in the SIMD instruction include headers of M1 Mac systems (macOS Big Sur and macOS Monterey), there are some incompatabilities with the ogre1.9 source files. The compiled state has been fully validated via the ROS RViz robotics visualization tool.

### DIFF
--- a/Formula/ogre1.9.rb
+++ b/Formula/ogre1.9.rb
@@ -26,63 +26,71 @@ class Ogre19 < Formula
 
   conflicts_with "ogre", because: "differing version of the same formula"
 
-  patch do
-    url "https://gist.github.com/NikolausDemmel/2b11d1b49b35cd27a102/raw/bf4a4d16020821218f73db0d56aa111ab2fde679/fix-1.9-HEAD.diff"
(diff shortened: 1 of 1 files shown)


scpeters:
thanks for the submission; hopefully this will fix #1628

scpeters:
There are currently a lot of patches on this formula, and it may be a bit unwieldy to work with them all, but I'd prefer to make smaller changes to this formula if possible. I analyzed the patch in [your gist](https://raw.githubusercontent.com/zarathustr/sinbad-ogre/master/mac-m1-full.patch), and I believe I have captured the new material in https://github.com/scpeters/ogre/commit/c39b6df49618773357da0b54437f351b57168d12 and that this patch could be added to this formula in a more concise way as in https://github.com/scpeters/homebrew-simulation/commit/27fae88e23d208102c60df519c43878b58ec5cff

To that end, I will open an alternative pull request with this more concise approach

Issue #1665, opened 2021-11-11
gazebo9,11 bottles broken by dartsim 6.12.1


scpeters:
The minor version of dartsim in homebrew-core was recently upgraded from 6.11.1 to 6.12.1 in https://github.com/Homebrew/homebrew-core/pull/89141, which has broken our gazebo bottles:

* [![Build Status](https://build.osrfoundation.org/buildStatus/icon?job=gazebo-install-one_liner-homebrew-amd64&build=2234)](https://build.osrfoundation.org/job/gazebo-install-one_liner-homebrew-amd64/2234/) https://build.osrfoundation.org/job/gazebo-install-one_liner-homebrew-amd64/2234/

We should open separate pull requests to bump the revision for gazebo9 and gazebo11 and rebuild the bottles. To do this, on a new branch run the following command for gazebo11:

~~~
brew bump-revision gazebo11 --message="for dartsim"
~~~

push the branch, open a pull request, and comment `build bottle`.

Then do the same for gazebo9.

nkalupahana:
^ can confirm, this has broken our builds that use gazebo11: https://github.com/nkalupahana/ros2-foxy-macos/runs/4176692550?check_suite_focus=true

nkalupahana:
All fixed for me!

PR #1669, opened 2021-11-11
gazebo9 bump for new DART

j-rivero:
part of  #1665


j-rivero:
build bottle

j-rivero:
Somehow the [builder job](https://build.osrfoundation.org/job/generic-release-homebrew_triggered_bottle_builder/657/label=osx_catalina/console) is not producing any bottle. @scpeters do you have any hint about what could be happening?

scpeters:
something changed recently in homebrew-test-bot that will skip bottling a formula if any of its dependencies are not bottled for that version of macOS. Previously it would use bottles from older macOS distros. I believe simbody and ignition-math4 are blocking gazebo9

There is a special way to use test-bot with the `--keep-old` flag to build bottles for a new distro only without bumping a revision. We should make a new jenkins job for this that has a package name (or names) and a single macOS distro name to use (such as bigsur or catalina) that will be used to construct the jenkins node label to use (osx_bigsur or osx_catalina). I'll open an issue for this on release-tools

scpeters:
> something changed recently in homebrew-test-bot that will skip bottling a formula if any of its dependencies are not bottled for that version of macOS. Previously it would use bottles from older macOS distros. I believe simbody and ignition-math4 are blocking gazebo9
> 
> There is a special way to use test-bot with the `--keep-old` flag to build bottles for a new distro only without bumping a revision. We should make a new jenkins job for this that has a package name (or names) and a single macOS distro name to use (such as bigsur or catalina) that will be used to construct the jenkins node label to use (osx_bigsur or osx_catalina). I'll open an issue for this on release-tools

https://github.com/ignition-tooling/release-tools/issues/555

scpeters:
I'll build again, though I expect it to fail since we haven't built bottles yet from #1675  to see if https://github.com/Homebrew/homebrew-test-bot/pull/717 gives useful console output

build bottle

scpeters:
build bottle

scpeters:
build bottle

PR #1683, opened 2021-11-17
Bumps in garden : ign-garden

scpeters:
See https://github.com/ignition-tooling/release-tools/issues/545

### DIFF
--- a/Formula/ignition-garden.rb
+++ b/Formula/ignition-garden.rb
@@ -3,18 +3,11 @@ class IgnitionGarden < Formula
 
   desc "Collection of gazebo simulation software"
   homepage "https://github.com/ignitionrobotics/ign-garden"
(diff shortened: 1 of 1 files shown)


Formula/ignition-garden.rb
```diff
@@ -27,7 +20,7 @@ class IgnitionGarden < Formula
   depends_on "ignition-msgs8"
   depends_on "ignition-physics5"
   depends_on "ignition-plugin1"
-  depends_on "ignition-rendering6"
+  depends_on "ignition-rendering7"
   depends_on "ignition-sensors6"
```
chapulina, line 24: Oops, I think it's missing some dependency bumps

scpeters: yeah, I'm debugging some changes to the bump scripts

PR #1698, opened 2021-12-01
Apple M1 fixes

dlech:
These fixes (along with https://github.com/Homebrew/homebrew-core/pull/90205) are required to get gazebo11 running on Apple M1 hardware as used by the https://github.com/PX4/PX4-Autopilot project.

### DIFF
--- a/Formula/ignition-common3.rb
+++ b/Formula/ignition-common3.rb
@@ -26,6 +26,13 @@ class IgnitionCommon3 < Formula
     cmake_args = std_cmake_args
     cmake_args << "-DBUILD_TESTING=Off"
     cmake_args << "-DCMAKE_INSTALL_RPATH=#{rpath}"
+
+    if Hardware::CPU.arm?
+      # REVISIT: this works around a crash on Apple M1 processors
+      # https://github.com/osrf/homebrew-simulation/pull/1698#discussion_r774674536
+      cmake_args << "-DIGN_PROFILER_REMOTERY=Off"
+    end
+
     system "cmake", ".", *cmake_args
     system "make", "install"
   end


scpeters (approved):
thanks for identifying these fixes!

Formula/ignition-common3.rb
```diff
@@ -25,6 +25,7 @@ class IgnitionCommon3 < Formula
   def install
     cmake_args = std_cmake_args
     cmake_args << "-DBUILD_TESTING=Off"
+    cmake_args << "-DIGN_PROFILER_REMOTERY=Off"
```
scpeters, line 28: is there a build problem with the Remotery profiler on M1?

dlech: No, not a build problem but rather a crash at runtime. As explained in the commit message, it causes a crash because of

```
symbol not found in flat namespace '_rmtMetal_usGetTime'
```

I didn't dig deeper, but I have a feeling that this has to do with clang adding a `_` prefix to C function names. The "missing" function is a C function that is called from ObjC code. I didn't dig any deeper than that. I was hoping that the profiler wasn't something important and we could just go with the easy fix. But if that is not the case, we will have to troubleshoot farther.

scpeters: ok, if it only occurs on the M1 hardware, then can you make this line conditional on the architecture? The following line has some logic you may consider using:

* https://github.com/Homebrew/homebrew-core/blob/4ca871cbc3f1747d6083eaf649aff8cc207f8fce/Formula/unar.rb#L36

dlech: I have made the requested change.

Desartstudio: I have the same problem "symbol not found in flat namespace '_rmtMetal_usGetTime'" in Mac OS12.1 intel I9 not M1

Desartstudio: ok thanks

PR #1734, opened 2021-12-21
🏁 Dome EOL: deprecate and disable bottles

chapulina:
* Part of https://github.com/ignition-tooling/release-tools/issues/600

### DIFF
--- a/Formula/ignition-dome.rb
+++ b/Formula/ignition-dome.rb
@@ -16,6 +16,10 @@ class IgnitionDome < Formula
     sha256 cellar: :any, mojave:   "25f52cf2b4893b2e98c3a9da733c33c434aa6eb9565f34bead4fb4eb725d6854"
   end
 
+  disable! date: "2022-01-31", because: "is past end-of-life date"
+
+  deprecate! date: "2021-12-31", because: "is past end-of-life date"
+
   depends_on "cmake" => :build
   depends_on "ignition-cmake2"
   depends_on "ignition-common3"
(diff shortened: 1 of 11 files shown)


j-rivero (approved):

scpeters:
I need to check if we can still build bottles after this has been merged; we may need to revert until all the dome releases are done

Formula/ignition-dome.rb
```diff
@@ -16,6 +16,10 @@ class IgnitionDome < Formula
     sha256 cellar: :any, mojave:   "25f52cf2b4893b2e98c3a9da733c33c434aa6eb9565f34bead4fb4eb725d6854"
   end
 
+  disable! date: "2022-01-31", because: "is past end-of-life date"
```
scpeters, line 19: adding a `disable!` tag automatically deprecates immediately

chapulina: * So you're saying that the date isn't respected? I was following what we did for Blueprint: https://github.com/osrf/homebrew-simulation/pull/1242

chapulina: * And now I see that was reverted in https://github.com/osrf/homebrew-simulation/pull/1256 :upside_down_face:

chapulina:
> I need to check if we can still build bottles after this has been merged

* Ah now tracing back what we did for Blueprint, I think it may not work: https://github.com/ignitionrobotics/sdformat/pull/449#issuecomment-752300521

Formula/ignition-dome.rb
chapulina: https://github.com/osrf/homebrew-simulation/pull/1737

PR #1762, opened 2021-12-27
Bumps in garden: ign-gazebo7 use ign-math7, ign-common5, sdformat13 and ign-physics6

chapulina:
See https://github.com/ignition-tooling/release-tools/issues/574


Formula/ignition-gazebo7.rb
```diff
@@ -2,19 +2,19 @@ class IgnitionGazebo7 < Formula
   desc "Ignition Gazebo robot simulator"
   homepage "https://github.com/ignitionrobotics/ign-gazebo"
   url "https://github.com/ignitionrobotics/ign-gazebo.git", branch: "main"
-  version "6.999.999~0~20211205"
+  version "6.999.999~0~20211227"
```
j-rivero, line 5: Are we building new source code or just rebuilding the existing code with new dependencies? If the former, it is fine.

chapulina: The source code has changed slightly to adapt to new APIs in the dependencies

scpeters: this formula points to the `main` branch, so it always builds whatever is the latest commit on `main`. The version bump on this line ensures that it will be upgraded if people already have it installed

scpeters (approved):

Formula/ignition-gazebo7.rb
j-rivero: > this formula points to the `main` branch, so it always builds whatever is the latest commit on `main`. The version bump on this line ensures that it will be upgraded if people already have it installed

I see, what was named at some point at a live package. If this is the case, I think it would be better to version it with something like `6.999.999~0~main` to reflect what users really get. Bumping 0 when we need to force a new version, like now. The date has very little to do with the reality of the code.

PR #1772, opened 2021-12-31
README: bottle removal docs and remove dome badges

scpeters:
This removes the dome badges and adds documentation for disabling broken bottles.


chapulina (approved):
Added the list cleanup as a step on the EOL template: https://github.com/ignition-tooling/release-tools/pull/616

PR #1776, opened 2022-01-05
ignition-math6,7: depend on pybind11 and swig

chapulina:
The goal is to enable testing with these dependencies on our CI.

If this is acceptable, I think the next step would be to add `--with-pybind11` and `--with-swig` to `PROJECT_ARGS` here:

https://github.com/ignition-tooling/release-tools/blob/master/jenkins-scripts/lib/project-default-devel-homebrew-amd64.bash#L77


scpeters:
brew doesn't really support optional dependencies very well anymore. I think we're better off adding these as regular dependencies. I'd like to test CI first though. The easiest way to do that is with a `ci_matching_branch/`, so I'll try creating an identical branch with that name

scpeters:
> brew doesn't really support optional dependencies very well anymore. I think we're better off adding these as regular dependencies. I'd like to test CI first though. The easiest way to do that is with a `ci_matching_branch/`, so I'll try creating an identical branch with that name

testing with https://github.com/ignitionrobotics/ign-math/pull/361

scpeters:
I see a cmake warning and several failed scripting tests, though the failed tests don't show up in jenkins because they don't generate junit files

* [Build Status](https://build.osrfoundation.org/job/ignition_math-ci-pr_any-homebrew-amd64/1476/) https://build.osrfoundation.org/job/ignition_math-ci-pr_any-homebrew-amd64/1476/
* https://build.osrfoundation.org/job/ignition_math-ci-pr_any-homebrew-amd64/1476/clang/

~~~
The following tests FAILED:
	119 - PID_TEST.py (Failed)
	139 - Angle_TEST.rb (Failed)
	140 - GaussMarkovProcess_TEST.rb (Failed)
	141 - Rand_TEST.rb (Failed)
	142 - Vector2_TEST.rb (Failed)
	143 - Vector3_TEST.rb (Failed)
	144 - Vector4_TEST.rb (Failed)
	145 - ruby_TEST.rb (Failed)
~~~

chapulina:
> I see a cmake warning and several failed scripting tests

Ah that's unfortunate, I was hoping it would just work. I won't have time to look into those, we could ticket an issue to see if someone volunteers.

scpeters:
closed in favor of #1793

PR #1786, opened 2022-01-14
dome: add disable! date 2 months after EOL

scpeters:
Part of https://github.com/ignition-tooling/release-tools/issues/600.

PR #1788, opened 2022-01-14
Disable old versions of gazebo in 2022-02-28

scpeters:
match disable dates in #1786


chapulina (approved):
I think we're missing a deprecation + disable for Gazebo 7

scpeters:
> I think we're missing a deprecation + disable for Gazebo 7

these formulae were already deprecated. I'll add new deprecations in a separate PR

PR #1793, opened 2022-01-26
ignition-math[6,7]: depend on pybind11

scpeters:
Alternative to #1776 without swig. Pairs with https://github.com/ignitionrobotics/ign-math/pull/361.


Formula/ignition-math6.rb
```diff
@@ -16,6 +16,7 @@ class IgnitionMath6 < Formula
   depends_on "doxygen" => :build
   depends_on "eigen"
   depends_on "ignition-cmake2"
+  depends_on "pybind11"
```
j-rivero, line 19:
```suggestion
  depends_on "pybind11" => :build
```
Is pybind11 a build dependency only? Looking into the debian packaging seems so to me.

scpeters: the `:build` flag ensures that a dependency is installed when building bottles but not when a bottle is installed. In order to ensure that we have `pybind11` installed in our CI, we need to upgrade our CI scripts. I will accept this change and open a PR to fix the release-tools scripts accordingly

PR #1814, opened 2022-02-18
ignition-gazebo6 6.6.0~pre1

osrfbuild:


j-rivero:
I was playing with the new python bindings packaging. We can probably skip this one, or build. @scpeters up to you, not sure if there are consequences if we just skip it.

scpeters:
yeah we can skip it for now

PR #1823, opened 2022-02-26
ogre2.2: patches for M1 arm processors

iche033:
Signed-off-by: Ian Chen <ichen@openrobotics.org>

I applied the patches mentioned in https://github.com/osrf/homebrew-simulation/issues/1628#issuecomment-1039718704 with a slight tweak to the first patch.

After this I was able to install ogre2.2, build and run the `simple_demo` example in ign-rendering with Metal backend.

simple_demo_m1_ign-rendering


iche033:
I also had to add an extra `"-DCMAKE_MACOSX_RPATH=FALSE",` cmake arg for ign-gazebo gui to load ogre libraries via dlopen otherwise I get `security policy does not allow @ path expansion` msgs.  I didn't add that change to this PR as I'm not sure if there are other better ways to around this.

Kakcalu13:
Do you know if this will apply to ultra M1 too?

iche033:
> Do you know if this will apply to ultra M1 too?

no I don't know. I don't have an m1 ultra to test this.

Kakcalu13:
> > Do you know if this will apply to ultra M1 too?
> 
> no I don't know. I don't have an m1 ultra to test this.

Sorry lol I don't know why I asked dumb question. I'm seriously considering to buy ultra M1. I am more than happy to test for you guys once I get it. By the time I get it, should I create a separate issue?

iche033:
I would suggest creating an issue once you are able to test this. You'll then be able to provide more info such as build errors and console output in the case that it does not work.

Kakcalu13:
build bottle
brew-bot-tag: build_m1_patch

Kakcalu13:
I have a quick question if you don't mind about this issue 👼,

How do you get this working?

Here is what I did:
1) cd /opt/homebrew/Library/Taps/osrf/homebrew-simulation/Formula
2) nano ogre2.2.rb
3) Copy/paste from the changed file on the issue here
4) Save and exit
5) brew tap osrf/simulation

Is this how you do this?

scpeters:
currently these patches will be applied for all architectures, so we should bump the revision and rebuild bottles

nello:
I've tried to apply these patches, and believe I have done so successfully. Still doesn't compile on my M1 Macbook Pro though.
```
==> cmake .. -DOGRE_LIB_DIRECTORY=lib/OGRE-2.2 -DOGRE_BUILD_LIBS_AS_FRAMEWORKS=OFF -DOGRE_FULL_RPATH
==> make install
Last 15 lines from ~/Library/Logs/Homebrew/ogre2.2/02.make:
      Ogre::ZipArchive::open(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, bool) in OgreZip.cpp.o
  "_zzip_file_read", referenced from:
      Ogre::ZipDataStream::read(void*, unsigned long) in OgreZip.cpp.o
  "_zzip_seek", referenced from:
      Ogre::ZipDataStream::skip(long) in OgreZip.cpp.o
      Ogre::ZipDataStream::seek(unsigned long) in OgreZip.cpp.o
  "_zzip_strerror_of", referenced from:
      Ogre::ZipDataStream::read(void*, unsigned long) in OgreZip.cpp.o
  "_zzip_tell", referenced from:
      Ogre::ZipDataStream::tell() const in OgreZip.cpp.o
ld: symbol(s) not found for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [lib/macosx/libOgreMain.2.2.6.dylib] Error 1
make[1]: *** [OgreMain/CMakeFiles/OgreMain.dir/all] Error 2
make: *** [all] Error 2
```

mjcarroll:
Does this set of patches apply to #1956 as well?  I only have an intel mac to test with.

iche033:
> Does this set of patches apply to #1956 as well? I only have an intel mac to test with.

I was able to build ogre2.3 on M1 without this patch

scpeters:
I've updated this branch and added some rpath fixes

scpeters:
build bottle

scpeters:
bottle upload is temporarily broken, wait to rebuild

PR #1859, opened 2022-04-03
Remove disabled formulae

scpeters:
The formulae from the Dome collection and numerous old gazebo classic formulae have been disabled for over a month. It's time to delete them.


Formula/gazebo7.rb
```diff
@@ -34,12 +34,6 @@ class Gazebo7 < Formula
   depends_on "gdal" => :optional
```
chapulina, line 34: We should deprecate / disable / remove `gazebo7`

chapulina (approved):
🧹

Formula/gazebo7.rb
scpeters: I will deprecate it soon

PR #1865, opened 2022-04-06
Bumps in garden : ign-cmake3

methylDragon:
See https://github.com/ignition-tooling/release-tools/issues/685

### DIFF
--- a/Aliases/ignition-cmake3
+++ b/Aliases/ignition-cmake3
@@ -1,1 +1,0 @@
-../Formula/ignition-cmake2.rb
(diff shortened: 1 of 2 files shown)


chapulina (approved):

chapulina:
I think we need to merge this one for #1866 to work. This doesn't break anything, so merging!

scpeters:
> I think we need to merge this one for #1866 to work. This doesn't break anything, so merging!

yes, this is safe to merge first

Issue #1884, opened 2022-04-13
Linking error when building ign-physics5 against dartsim@6.10.0


scpeters:
Something has changed recently that is causing linking errors in ignition-physics5. The following builds are both using https://github.com/ignitionrobotics/ign-physics/commit/0111a8f5033947abc9105ff6b7a7ad36f351a79c but were built at different times, suggesting that something in the build infrastructure or homebrew dependencies changed in between these builds:

* [![Build Status](https://build.osrfoundation.org/buildStatus/icon?job=ignition_physics-ci-ign-physics5-homebrew-amd64&build=12)](https://build.osrfoundation.org/view/ign-fortress/job/ignition_physics-ci-ign-physics5-homebrew-amd64/12/) https://build.osrfoundation.org/view/ign-fortress/job/ignition_physics-ci-ign-physics5-homebrew-amd64/12/
* [![Build Status](https://build.osrfoundation.org/buildStatus/icon?job=ignition_physics-ci-ign-physics5-homebrew-amd64&build=13)](https://build.osrfoundation.org/view/ign-fortress/job/ignition_physics-ci-ign-physics5-homebrew-amd64/13/) https://build.osrfoundation.org/view/ign-fortress/job/ignition_physics-ci-ign-physics5-homebrew-amd64/13/

The failing build has the following console output:

~~~
ld: library not found for -lTINYXML2_LIBRARY_/usr/local/Cellar/tinyxml2/9.0.0/lib/libtinyxml2.dylib-NOTFOUND
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [lib/libMockDoublePendulum.dylib] Error 1
make[1]: *** [test/plugins/CMakeFiles/MockDoublePendulum.dir/all] Error 2
...
ld: library not found for -lTINYXML2_LIBRARY_/usr/local/Cellar/tinyxml2/9.0.0/lib/libtinyxml2.dylib-NOTFOUND
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [lib/libignition-physics5-dartsim-plugin.5.1.0.dylib] Error 1
make[1]: *** [dartsim/CMakeFiles/ignition-physics5-dartsim-plugin.dir/all] Error 2
~~~

Both builds are using the `dartsim@6.10.0` bottle from our tap. I have reproduced this failure when building against `dartsim@6.10.0` on my local machine, but when building against the `dartsim` bottle from homebrew-core, it links successfully.

[...]

PR #1885, opened 2022-04-13
ignition-physics5: don't use dartsim fork

scpeters:
Switch from dartsim@6.10.0 back to upstream dartsim to see if that fixes linking problems documented in #1884.


scpeters:
build bottle

Formula/ignition-physics5.rb
```diff
@@ -15,7 +15,7 @@ class IgnitionPhysics5 < Formula
   depends_on "cmake" => :build
 
   depends_on "bullet"
-  depends_on "dartsim@6.10.0"
+  depends_on "dartsim"
```
azeey, line 18: This will install the latest version right (6.12.1)?

azeey (approved):

Formula/ignition-physics5.rb
scpeters: correct https://github.com/Homebrew/homebrew-core/blob/master/Formula/dartsim.rb#L4

PR #1932, opened 2022-06-08
ignition-utils2: Rename CMake project to gz-utils2

chapulina:
* Part of https://github.com/gazebo-tooling/release-tools/issues/718
* Goes with https://github.com/gazebosim/gz-utils/pull/57


chapulina:
Merging so we can progress with the renaming. The test changes will only be verified when we do a pre-release.

PR #1933, opened 2022-06-08
ignition-math7: Rename CMake project to gz-math7

chapulina:
* Part of https://github.com/gazebo-tooling/release-tools/issues/718
* Goes with https://github.com/gazebosim/gz-math/pull/420


chapulina:
Merging so we can progress with the renaming. The test changes will only be verified when we do a pre-release.

PR #1950, opened 2022-06-30
gz-tools2: don't install bin/ign symlink

scpeters:
This allows garden to be installed side-by-side with citadel and fortress.

Though it currently errors if I try to run either command:

~~~
$ gz
Configuration error: File [/usr/local/share/ignition/transport12.yaml] contains a command already registered by ignition-transport.
$ ign
Configuration error: File [/usr/local/share/ignition/transport12.yaml] contains a command already registered by ignition-transport.
~~~


scpeters:
resolved upstream in https://github.com/gazebosim/gz-tools/pull/109

Issue #1860, opened 2022-04-03
Rebuild bottles once protobuf 3.20.0 is merged


scpeters:
Any bottles that depend on protobuf will be invalidated when https://github.com/Homebrew/homebrew-core/pull/98436 is merged. Similar to https://github.com/osrf/homebrew-simulation/issues/1728

* [ ] Remove all broken bottles (`gazebo*`, `ignition-msgs*` and their dependents) 
* [ ] Rebuild ignition-msgs*
* [ ] etc

scpeters:
I should make a template for this type of issue

scpeters:
protobuf 3.20.1 is now under review in https://github.com/Homebrew/homebrew-core/pull/99843

scpeters:
protobuf 21.2 was finally merged in https://github.com/Homebrew/homebrew-core/pull/105712, so this may proceed

scpeters:
all bottles have been rebuilt

PR #1959, opened 2022-07-16
Rebuild fortress bottles after protobuf bump.

scpeters:
Part of https://github.com/osrf/homebrew-simulation/issues/1860.

Since `ignition-msgs8` had already been rebuilt, its dependents could be bumped with:

~~~
for f in $(brew uses ignition-msgs8)
do
  echo $f
  brew bump-revision --message="rebuild after protobuf" $f
done
~~~


scpeters:
build bottle

PR #1960, opened 2022-07-16
gz-msgs5: rebuild after protobuf bump

scpeters:
Part of https://github.com/osrf/homebrew-simulation/issues/1860. Unreleased patch from https://github.com/gazebosim/gz-msgs/pull/242 needed to fix compilation.


scpeters:
build bottle

PR #1971, opened 2022-07-26
ogre2.3: use CMake args to build components

mjcarroll:
There were a few components that weren't being correctly generated without explicitly enabling them.


scpeters:
build bottle

scpeters (approved):

PR #1976, opened 2022-07-29
gz-common5: depend on libcsv

mjcarroll:
Required for https://github.com/gazebosim/gz-common/pull/402


scpeters (approved):

scpeters:
required by https://github.com/gazebosim/gz-cmake/pull/290

chapulina:
We should revert this now that we're no longer going to depend on `libcsv` since it's LGPL.

scpeters:
> We should revert this now that we're no longer going to depend on `libcsv` since it's LGPL.

reverted in https://github.com/osrf/homebrew-simulation/pull/1977

PR #2039, opened 2022-09-08
gz-tools2: fix default data path

azeey:
This avoids the need for setting `GZ_CONFIG_PATH` for gz libraries installed via homebrew.


azeey:
Using `inreplace` in https://github.com/osrf/homebrew-simulation/pull/2039/commits/d9b2c844abca21c13dfb0b8f9065eccd098b7704 instead of a patch as that is the preferred method for this kind of fix.

scpeters:
looks great! We should bump the revision number and build new bottles

~~~
brew bump-revision gz-tools2
~~~

I can take care of this if you prefer

scpeters:
build bottle

scpeters (approved):

PR #2040, opened 2022-09-09
ignition-tools: fix default data path

scpeters:
Similar to https://github.com/osrf/homebrew-simulation/pull/2039.

This avoids the need for setting `IGN_CONFIG_PATH` for `ign` libraries installed via homebrew.


scpeters:
build bottle

PR #2063, opened 2022-09-28
gz-launch6 6.0.0

osrfbuild:

PR #2065, opened 2022-09-28
Bump rev in gz-launch6/gz-sim7 to build bottle

j-rivero:
Forgot to call build bottle in #2063 so bumping rev to get a new one


j-rivero:
@osrfbuild build bottle

j-rivero:
@osrfbuild build bottle

PR #2159, opened 2022-12-04
gz-harmonic: alias to gz-garden

scpeters:
There is a harmonic bottle CI job since https://github.com/gazebo-tooling/release-tools/pull/857/ was merged, but it fails because there is no `gz-harmonic` formula:

* [Build Status](https://build.osrfoundation.org/job/ignition_harmonic-install_bottle-homebrew-amd64/4/) https://build.osrfoundation.org/job/ignition_harmonic-install_bottle-homebrew-amd64/4/

~~~
+ brew install --include-test ignition-harmonic
Warning: No available formula with the name "ignition-harmonic".
~~~

So let's just make an alias for now.

### DIFF
--- a/Aliases/gz-harmonic
+++ b/Aliases/gz-harmonic
@@ -1,0 +1,1 @@
+../Formula/gz-garden.rb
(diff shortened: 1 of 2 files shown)


scpeters:
merging to fix CI

PR #2194, opened 2023-02-03
gz-physics6 6.3.0

osrfbuild:


azeey:
build bottle

scpeters:
good call removing the patch

build bottle

PR #2196, opened 2023-02-03
gz-transport13: depend on gz-msgs10

j-rivero:
Remove the alias, copy the transport12 formula and bump dependency on msgs to version 10. Update all references to 12 to 13.

### DIFF
--- a/Formula/gz-transport13.rb
+++ b/Formula/gz-transport13.rb
@@ -1,4 +1,4 @@
-class Gztransport13 < Formula
+class GzTransport13 < Formula
   desc "Transport middleware for robotics"
   homepage "https://gazebosim.org"
   url "https://github.com/gazebosim/gz-transport.git", branch: "main"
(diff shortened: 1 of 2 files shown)


caguero (approved):
Not sure if it's needed but I've seen that we create an alias for the next version (`transport14`) in similar PRs...

scpeters:
> Not sure if it's needed but I've seen that we create an alias for the next version (`transport14`) in similar PRs...

it's nice to do but not required

Formula/gz-transport13.rb
```diff
@@ -0,0 +1,71 @@
+class Gztransport13 < Formula
```
scpeters, line 1: brew complaint: 

~~~
Expected to find class GzTransport13, but only found: Gztransport13.
~~~

caguero: [5f8c284](https://github.com/osrf/homebrew-simulation/pull/2196/commits/5f8c2842ce862bab4545de8af59f6157d1e225cc)

PR #2213, opened 2023-02-16
gz-math7: use python@3.11, revision bump

mjcarroll:


mjcarroll:
build bottle

scpeters:
we might need to update sdformat13 and gz-garden as well, but let's see if this works

scpeters:
build bottle

scpeters:
build bottle

PR #2214, opened 2023-02-16
sdf13: use python@3.11, revision bump

mjcarroll:


scpeters:
this needs to be done at the same time as #2213, so I've applied these changes there. I didn't see this PR in time or I would have cherry-picked your commit, sorry

Issue #1990, opened 2022-08-10
Bottle build may fail if target branch receives commits


scpeters:
The [generic-release-homebrew_triggered_bottle_builder](https://build.osrfoundation.org/job/generic-release-homebrew_triggered_bottle_builder) job builds bottles for separate configurations with jobs that run separately. Currently, the first build is for `osx_catalina`, followed by `osx_bigsur`. Each configuration job attempts to check out `refs/remotes/origin/pr/*/merge` for the corresponding pull request.

If there is a commit to the target branch after the first configuration has started and before the second, the second job may fail with an error message like the following:

~~~
Checking out Revision dc6001d45a119903288bc3280342d90083d8fefd (refs/remotes/origin/pr/1987/merge)
 > git config core.sparsecheckout # timeout=10
 > git checkout -f dc6001d45a119903288bc3280342d90083d8fefd # timeout=10
FATAL: Could not checkout dc6001d45a119903288bc3280342d90083d8fefd
hudson.plugins.git.GitException: Command "git checkout -f dc6001d45a119903288bc3280342d90083d8fefd" returned status code 128:
stdout: 
stderr: fatal: reference is not a tree: dc6001d45a119903288bc3280342d90083d8fefd

	at org.jenkinsci.plugins.gitclient.CliGitAPIImpl.launchCommandIn(CliGitAPIImpl.java:2671)
	at org.jenkinsci.plugins.gitclient.CliGitAPIImpl.access$1100(CliGitAPIImpl.java:84)
	at org.jenkinsci.plugins.gitclient.CliGitAPIImpl$9.execute(CliGitAPIImpl.java:2989)
~~~

[![Build Status](https://build.osrfoundation.org/buildStatus/icon?job=generic-release-homebrew_triggered_bottle_builder%2Flabel%3Dosx_bigsur&build=901)](https://build.osrfoundation.org/job/generic-release-homebrew_triggered_bottle_builder/label=osx_bigsur/901/) https://build.osrfoundation.org/job/generic-release-homebrew_triggered_bottle_builder/901/

PR #2224, opened 2023-03-03
gz-rendering7 7.3.0

osrfbuild:


iche033:
build bottle

scpeters:
I think the builder failed due to https://github.com/osrf/homebrew-simulation/issues/1990 since #2223 was merged while the builder job was still running. It should be fine to restart it and let it try again

scpeters:
build bottle

PR #2240, opened 2023-04-21
Add issue template for broken bottles

scpeters:
Also request `brew config` in new issues.


scpeters:
I've noticed some bottles broken by ffmpeg 6.0, so I'm updating our issue template and I will try using it to report this failure and see if it helps with the resolution process. I will merge this without review in order to test creating the issue with this template.

PR #2273, opened 2023-06-02
Update CMAKE_INSTALL_RPATH for gz-garden dependencies

srmainwaring:
Update the `CMAKE_INSTALL_RPATH` to include `/opt/homebrew/lib` which is the brew library location for Apple M1 machines.

Fixes: https://github.com/gazebosim/gz-sim/issues/1990

Tested on macOS Ventura, Xcode and CommandLineTools 14.3.1. MacBookPro M1.

### DIFF
--- a/Formula/gz-common5.rb
+++ b/Formula/gz-common5.rb
@@ -30,7 +30,7 @@ class GzCommon5 < Formula
   def install
     cmake_args = std_cmake_args
     cmake_args << "-DBUILD_TESTING=Off"
-    cmake_args << "-DCMAKE_INSTALL_RPATH=#{rpath}"
+    cmake_args << "-DCMAKE_INSTALL_RPATH=#{rpath};/opt/homebrew/lib"
 
     # Use build folder
     mkdir "build" do
--- a/Formula/gz-fuel-tools8.rb
+++ b/Formula/gz-fuel-tools8.rb
@@ -26,7 +26,7 @@ class GzFuelTools8 < Formula
   def install
     cmake_args = std_cmake_args
     cmake_args << "-DBUILD_TESTING=Off"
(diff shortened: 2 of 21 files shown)


scpeters:
apologies for neglecting this for so long; I don't have easy access to an M1 machine for testing this, but I've found some time at the moment and am looking into the RPATH issues on my personal M1 machine

scpeters:
I think the approach in this PR is a bit too broad; I'd like to be more targeted with the RPATH.

For example, I found an issue with the CLI executable in gz-plugin2 (https://github.com/osrf/homebrew-simulation/issues/2403#issuecomment-1716888880), and I've identified two ways to add the needed relative path from the libexec subfolder to `lib`:

* add the needed relative path in the homebrew formula: https://github.com/osrf/homebrew-simulation/pull/2404
* specify the RPATH with `APPLE`-specific cmake code in the upstream repo: https://github.com/gazebosim/gz-plugin/commit/c5d69a40ae73d1de0f5f9ab1ea5f734fe0d4bba6

I'm leaning towards the first option for now. I think any package with executables or shared libraries that aren't in sibling folders of `lib` will need this treatment.

scpeters:
please take a look at https://github.com/osrf/homebrew-simulation/pull/2409 for an alternative approach to fixing rpath for garden packages other than gz-gui7 (which has issues I haven't figured out yet)

scpeters:
I think I have resolved the RPATH issues for garden with https://github.com/osrf/homebrew-simulation/pull/2409 and https://github.com/osrf/homebrew-simulation/pull/2410

scpeters:
I'm going to close this because I believe it has been fixed by #2409 and #2410

PR #2275, opened 2023-06-08
Remove broken bottles, depend on protobuf@21

scpeters:
Part of https://github.com/osrf/homebrew-simulation/issues/2274.


scpeters:
from local testing I had to disable bottles for all dependents of gz-msgs

PR #2276, opened 2023-06-08
Rebuild garden bottles, add protobuf@21 env variables

scpeters:
Part of #2274


scpeters:
build bottle

scpeters:
apply patches first, then rebuild

PR #2322, opened 2023-07-28
gz-launch7: fix version

azeey:
See https://github.com/gazebo-tooling/release-tools/issues/976

### DIFF
--- a/Formula/gz-launch7.rb
+++ b/Formula/gz-launch7.rb
@@ -2,7 +2,7 @@ class GzLaunch7 < Formula
   desc "Launch libraries for robotics applications"
   homepage "https://github.com/gazebosim/gz-launch"
   url "https://github.com/gazebosim/gz-launch.git", branch: "main"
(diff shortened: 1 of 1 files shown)


Formula/gz-launch7.rb
```diff
@@ -2,7 +2,7 @@ class GzLaunch7 < Formula
   desc "Launch libraries for robotics applications"
   homepage "https://github.com/gazebosim/gz-launch"
   url "https://github.com/gazebosim/gz-launch.git", branch: "main"
-  version "7.999.999~0~20221117"
+  version "6.999.999~0~20230728"
```
azeey, line 5: @scpeters would this cause problems? I think 6.999.999 is the correct value, but the old one had 7.999.999.

scpeters: yeah, it should be fine since there are no bottles built for this yet

scpeters (approved):

PR #2353, opened 2023-08-29
gz-msgs10 10.0.0~pre1

osrfbuild:


azeey:
@osrf-jenkins build bottle

PR #2369, opened 2023-08-31
gz-harmonic 1.0.0~pre1

osrfbuild:


scpeters:
I think gz-harmonic has nothing to install. We should install a `gazebodistro` folder like gz-garden has:

* https://github.com/gazebosim/gz-garden/tree/main/gazebodistro
* https://github.com/gazebosim/gz-garden/blob/main/CMakeLists.txt#L57

j-rivero:
> I think gz-harmonic has nothing to install. We should install a `gazebodistro` folder like gz-garden has:

I go for it

j-rivero:
pre2 in https://github.com/osrf/homebrew-simulation/pull/2373

PR #2393, opened 2023-09-07
Testing don't merge gz-harmonic 1.0.0~pre3

osrfbuild:


scpeters:
this was only for testing https://github.com/gazebo-tooling/release-tools/pull/1007

Issue #2394, opened 2023-09-08
Need to rebuild bottles for protobuf 24.3


scpeters:
I noticed that https://github.com/Homebrew/homebrew-core/pull/141741 may be merged soon. If this breaks our bottles, we should be ready to quickly remove broken bottles

Confirmed with [![Build Status](https://build.osrfoundation.org/buildStatus/icon?job=ignition_msgs9-install_bottle-homebrew-amd64&build=318)](https://build.osrfoundation.org/view/ign-garden/job/ignition_msgs9-install_bottle-homebrew-amd64/318/) https://build.osrfoundation.org/view/ign-garden/job/ignition_msgs9-install_bottle-homebrew-amd64/318/

Please supply your brew configuration:

### Brew configuration

~~~
$ brew config
HOMEBREW_VERSION: 4.1.10
ORIGIN: https://github.com/Homebrew/brew
HEAD: 12c8778af976269344f3c0b914243e178fc4ffbd
Last commit: 2 days ago
Core tap origin: https://github.com/Homebrew/homebrew-core
Core tap HEAD: efba59ef1a9ae0183198311a72f0ed07eac93ee3
Core tap last commit: 20 minutes ago
Core tap branch: master
Core tap JSON: 08 Sep 11:12 UTC
HOMEBREW_PREFIX: /usr/local
HOMEBREW_CASK_OPTS: []
HOMEBREW_GIT_EMAIL: osrfbuild@osrfoundation.org
HOMEBREW_GIT_NAME: OSRF Build Bot
HOMEBREW_MAKE_JOBS: 12
HOMEBREW_NO_AUTO_UPDATE: set
HOMEBREW_NO_INSTALL_FROM_API: set
HOMEBREW_SORBET_RUNTIME: set
HOMEBREW_UPDATE_TO_TAG: set
Homebrew Ruby: 2.6.10 => /usr/local/Homebrew/Library/Homebrew/vendor/portable-ruby/2.6.10_1/bin/ruby
CPU: dodeca-core 64-bit kabylake
Clang: 13.0.0 build 1300
Git: 2.42.0 => /usr/local/bin/git
Curl: 7.87.0 => /usr/bin/curl
macOS: 11.7.7-x86_64
CLT: 13.2.0.0.1.1638488800
Xcode: N/A
~~~

### Brew linkage test result

~~~
$ brew linkage --test ignition-msgs9
Broken dependencies:
[...]

PR #2400, opened 2023-09-12
Fortress: patch physics and rebuild bottles

scpeters:
This adds a patch for gz-physics5 from https://github.com/gazebosim/gz-physics/pull/537 so that the gz-sim6 test on monterey can be enabled.

This also rebuilds the fortress bottles that depend on protobuf as part of #2394.


scpeters:
build bottle

scpeters:
build bottle

Issue #2403, opened 2023-09-13
ARM: gz-plugin2 CLI executable has broken RPATH


scpeters:
## Description:

The executable in `gz-plugin2`'s `libexec` folder is not able to load the `libgz-plugin2-loader.2.dylib` library when using an ARM processor with homebrew installed in `/opt/homebrew`: 

~~~
$(brew --prefix)/opt/gz-plugin2/libexec/gz/plugin2/gz-plugin
dyld[90133]: Library not loaded: '@rpath/libgz-plugin2-loader.2.dylib'
  Referenced from: '/opt/homebrew/Cellar/gz-plugin2/2.0.1/libexec/gz/plugin2/gz-plugin'
  Reason: tried: '/opt/homebrew/Cellar/gz-plugin2/2.0.1/libexec/gz/plugin2/../lib/libgz-plugin2-loader.2.dylib' (no such file), '/opt/homebrew/Cellar/gz-plugin2/2.0.1/libexec/gz/plugin2/../lib/libgz-plugin2-loader.2.dylib' (no such file), '/usr/local/lib/libgz-plugin2-loader.2.dylib' (no such file), '/usr/lib/libgz-plugin2-loader.2.dylib' (no such file)
Abort trap: 6
~~~

## Please supply your brew configuration:

~~~
$ brew config
HOMEBREW_VERSION: 4.1.11-2-geacdf7a
ORIGIN: https://github.com/Homebrew/brew
HEAD: eacdf7aa60a1a8a64acc71853ec5559cfd2329c3
Last commit: 22 hours ago
Core tap origin: https://github.com/Homebrew/homebrew-core
Core tap HEAD: 6ff75ff4ca4500b187f786536ed90867fa1db3d7
Core tap last commit: 7 hours ago
Core tap branch: master
Core tap JSON: 12 Sep 19:39 UTC
HOMEBREW_PREFIX: /opt/homebrew
HOMEBREW_CASK_OPTS: []
HOMEBREW_MAKE_JOBS: 8
HOMEBREW_SORBET_RUNTIME: set
Homebrew Ruby: 2.6.10 => /System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/bin/ruby
CPU: octa-core 64-bit arm_firestorm_icestorm
Clang: 14.0.0 build 1400
Git: 2.37.1 => /Library/Developer/CommandLineTools/usr/bin/git
Curl: 8.1.2 => /usr/bin/curl
[...]

PR #2404, opened 2023-09-13
gz-plugin: fix rpath for libexec/gz/plugin2

scpeters:
Add an rpath entry with a relative path from the libexec subfolder containing the CLI executable to lib.

Fixes https://github.com/osrf/homebrew-simulation/issues/2403#issuecomment-1716888880.

This could alternatively be fixed in gz-plugin with something like https://github.com/gazebosim/gz-plugin/commit/c5d69a40ae73d1de0f5f9ab1ea5f734fe0d4bba6.

For reference, I copied the implementation for this from the [arrayfire](https://github.com/Homebrew/homebrew-core/blob/master/Formula/a/arrayfire.rb#L36-L51) formula in homebrew-core.


scpeters:
build bottle

gmerlino:
Should the same rpath fixes be applied to Harmonic-related formulas? @scpeters

scpeters:
> Should the same rpath fixes be applied to Harmonic-related formulas? @scpeters

see #2413

PR #2409, opened 2023-09-15
Garden: fix physics, rendering, transport rpath issues and add tests

scpeters:
This adds custom CMAKE_INSTALL_RPATH entries for
formulae with binaries installed to nested subfolders, such as <prefix>/libexec/gz/transport12 and
<prefix>/lib/gz-physics-6/engine-plugins.
This should fix library loading issues with ARM CPUs. This adds tests as well.

* gz-rendering: patch for ogre2 rpath

This is a follow-up to #2404.

I'm having trouble with gz-gui, but have fixed the other packages.

### DIFF
--- a/Formula/gz-launch6.rb
+++ b/Formula/gz-launch6.rb
@@ -4,7 +4,7 @@ class GzLaunch6 < Formula
   url "https://osrf-distributions.s3.amazonaws.com/gz-launch/releases/gz-launch-6.0.0.tar.bz2"
   sha256 "e59e988c8a454cacc9f8f5727d1ab0d2c7fc8476083ec96065a0f89913957b0a"
   license "Apache-2.0"
-  revision 11
+  revision 12
 
   head "https://github.com/gazebosim/gz-launch.git", branch: "gz-launch6"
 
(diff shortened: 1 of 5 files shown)


scpeters:
build bottle

scpeters:
build bottle

scpeters:
build bottle

iche033:
what's the proper way to install and test this? I installed by doing `brew install --build-from-source ./gz-<lib_name>.rb` and then was able to launch `gz sim -v 4 -s shapes.sdf` but the gui plugins (in both gz-sim and gz-gui) complains about rpath issues.

scpeters:
> what's the proper way to install and test this? I installed by doing `brew install --build-from-source ./gz-<lib_name>.rb` and then was able to launch `gz sim -v 4 -s shapes.sdf` but when launching the gui, the gui plugins (in both gz-sim and gz-gui) complain about rpath issues.

I can't figure out how to fix the rpath issues for the gz-gui plugins, and I forgot one of the gz-sim gui plugin folders, so this PR is just about getting the gz-sim server running.

iche033:
> I can't figure out how to fix the rpath issues for the gz-gui plugins, and I forgot one of the gz-sim gui plugin folders, so this PR is just about getting the gz-sim server running.

ok got it. `gz sim -s` works for me.

iche033 (approved):

scpeters:
I've tracked down the issues with the gui plugins, and it will require further changes to gz-sim7, so I've redacted those changes from this pull request and will update it in a follow-up

PR #2410, opened 2023-09-18
Garden: rpath issues in gui plugins, ogre2.3

scpeters:
Follow-up to #2409, #2404.


scpeters:
build bottle

iche033:
nice, `gz sim -s` and `-g` both work for me without any rpath errors now!

iche033 (approved):

PR #2462, opened 2023-10-09
Bumps in ionic : ci_matching_branch/bump_ionic_gz-tools3

azeey:
See https://github.com/gazebo-tooling/release-tools/issues/1027


azeey:
We've decided not to bump gz-tools

PR #2463, opened 2023-10-09
gz-ionic: new formula in ionic

azeey:
See https://github.com/gazebo-tooling/release-tools/issues/1027


Formula/gz-ionic.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    venv = virtualenv_create(libexec, Formula["python@3.11"].opt_libexec/"bin/python")
+    %w[PyYAML vcstool].each do |pkg|
+      venv.pip_install pkg
+    end
+  end
+
+  test do
+    yaml_file = share/"gz/gz-ionic/gazebodistro/collection-ionic.yaml"
+    system libexec/"bin/vcs", "validate", "--input", yaml_file
```
scpeters, line 52: we should change this test if we are installing release notes instead of a yaml file

Formula/gz-ionic.rb
```diff
@@ -0,0 +1,54 @@
...
+  def install
+    mkdir "build" do
+      system "cmake", "..", *std_cmake_args
+      system "make", "install"
+    end
+
+    venv = virtualenv_create(libexec, Formula["python@3.11"].opt_libexec/"bin/python")
+    %w[PyYAML vcstool].each do |pkg|
+      venv.pip_install pkg
+    end
```
scpeters, line 47: this can be removed if we don't install and test the yaml

Formula/gz-ionic.rb
```diff
@@ -0,0 +1,54 @@
...
+  depends_on "gz-tools2"
+  depends_on "gz-transport14"
+  depends_on "gz-utils3"
+  depends_on "pkg-config"
+  depends_on "sdformat15"
+
+  resource "PyYAML" do
+    url "https://files.pythonhosted.org/packages/64/c2/b80047c7ac2478f9501676c988a5411ed5572f35d1beff9cae07d321512c/PyYAML-5.3.1.tar.gz"
+    sha256 "b8eac752c5e14d3eca0e6dd9199cd627518cb5ec06add0de9d32baeee6fe645d"
+  end
```
scpeters, line 36: this can be removed if we don't install and test the yaml

Formula/gz-ionic.rb
```diff
@@ -0,0 +1,54 @@
...
+  desc "Collection of gazebo simulation software"
+  homepage "https://github.com/gazebosim/gz-ionic"
+  url "https://github.com/gazebosim/gz-ionic.git", branch: "main"
+  version "0.999.999-0-20231016"
+  license "Apache-2.0"
+
+  head "https://github.com/gazebosim/gz-ionic.git", branch: "main"
+
+  depends_on "cmake" => :build
+  depends_on "python@3.11" => [:build, :test]
```
scpeters, line 13: this can be removed if we don't install and test the yaml

scpeters (changes requested):

azeey:
@scpeters I've addressed your feedback. I'm not sure why CI is failing though.

iche033:
the latest brew test error could be due to having a gz-ionic file in Aliases. Let me try removing that

scpeters (approved):

PR #2481, opened 2023-10-26
Fix linting problems with github tarball refs

j-rivero:
Seems like brew has changed the linting to avoid github archive paths not using refs/tags, see https://github.com/osrf/homebrew-simulation/actions/runs/6655393821/job/18085564469?pr=2480.

Let's try to make it happy.


j-rivero:
Merging to solve the testing of package creation with new centralized tarballs.

Issue #2483, opened 2023-11-02
Rebuild bottles after protobuf 25 is in homebrew-core


scpeters:
I expect all bottles that depend on gz-msgs* to break after protobuf 25 is merged in https://github.com/Homebrew/homebrew-core/pull/153117 (similar to #2339, #2394, #2455 in the past two months).

Please supply your brew configuration:

### Brew configuration

~~~
$ brew config
HOMEBREW_VERSION: 4.1.14
ORIGIN: https://github.com/Homebrew/brew
HEAD: affc4c01aada2c973b63e084e7696e896edf2b7b
Last commit: 7 days ago
Core tap origin: https://github.com/Homebrew/homebrew-core
Core tap HEAD: 8801654860a462b0efc21606d45fbd29cc090238
Core tap last commit: 17 minutes ago
Core tap branch: master
Core tap JSON: 30 Sep 11:12 UTC
HOMEBREW_PREFIX: /usr/local
HOMEBREW_CASK_OPTS: []
HOMEBREW_GIT_EMAIL: osrfbuild@osrfoundation.org
HOMEBREW_GIT_NAME: OSRF Build Bot
HOMEBREW_MAKE_JOBS: 12
HOMEBREW_NO_AUTO_UPDATE: set
HOMEBREW_NO_INSTALL_FROM_API: set
HOMEBREW_SORBET_RUNTIME: set
HOMEBREW_UPDATE_TO_TAG: set
Homebrew Ruby: 2.6.10 => /System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/bin/ruby
CPU: dodeca-core 64-bit kabylake
Clang: 15.0.0 build 1500
Git: 2.42.0 => /usr/local/bin/git
Curl: 8.1.2 => /usr/bin/curl
macOS: 13.6-x86_64
CLT: 15.0.0.0.1.1694021235
~~~

### Brew linkage test result

~~~
$ brew linkage --test gz-msgs9
Broken dependencies:
  /usr/local/opt/protobuf/lib/libprotobuf.24.4.0.dylib (protobuf)
~~~

### Checklist

- [X] Identify change in homebrew-core that broke the bottle https://github.com/Homebrew/homebrew-core/pull/153117
- [X] Identify list of affected formulae
    * gz-msgs*
    * gz-transport*
    * gz-fuel-tools*
    * gz-gui*
    * gz-sensors*
    * gz-sim*
    * gz-launch*
    * gazebo11
[...]

PR #2485, opened 2023-11-04
Rebuild harmonic bottles for protobuf 25

scpeters:
Part of #2483.

Pull request commits created with the following command:

~~~
for f in $(grep -L '^ *bottle do' $(grep -rnI depend.*gz gz-harmonic.rb | sed -e 's@.* "@@' -e 's@"@.rb@'))
do
    brew bump-revision --message="rebuild bottle" $f
done
~~~


scpeters:
build bottle

PR #2535, opened 2024-01-20
gz-math6: use python@3.12, fix bindings install

scpeters:
Part of https://github.com/gazebosim/gz-sim/issues/2249. Switch to python@3.12 since that version is used by `gdal` and is preferred by cmake `find_package` calls over 3.11.


scpeters:
build bottle

scpeters:
trying a different approach

Issue #2527, opened 2024-01-10
Rebuild bottles after protobuf 25.2 is merged


scpeters:
I expect all bottles using protobuf to break when the following PR is merged (similar to https://github.com/osrf/homebrew-simulation/issues/2493):

* https://github.com/Homebrew/homebrew-core/pull/159459

### Checklist

- [X] Identify change in homebrew-core that broke the bottle https://github.com/Homebrew/homebrew-core/pull/159459
- [X] Identify list of affected formulae
    * gz-msgs*
    * gz-transport*
    * gz-fuel-tools*
    * gz-gui*
    * gz-sensors*
    * gz-sim*
    * gz-launch*
    * gazebo11
- [x] Remove broken bottles: for each affected formula, run `brew bump-revision --remove-bottle-block`
    - [X]  https://github.com/osrf/homebrew-simulation/pull/2528
- [x] Rebuild broken bottles
    - [x] Harmonic
      - [x] https://github.com/osrf/homebrew-simulation/pull/2531
      - [x] https://github.com/osrf/homebrew-simulation/pull/2532
    - [x] Garden
      - [x] https://github.com/osrf/homebrew-simulation/pull/2533
    - [x] Fortress
      - [x] https://github.com/osrf/homebrew-simulation/pull/2536
      - [ ] https://github.com/osrf/homebrew-simulation/pull/2540
      - [x] https://github.com/osrf/homebrew-simulation/pull/2547
    - [x] Citadel
      - [x] https://github.com/osrf/homebrew-simulation/pull/2537
      - [x] https://github.com/osrf/homebrew-simulation/pull/2546
    - [x] gazebo11
      - [x] https://github.com/osrf/homebrew-simulation/pull/2538

PR #2536, opened 2024-01-20
Rebuild most fortress bottles

scpeters:
Part of https://github.com/osrf/homebrew-simulation/issues/2527.

Rebuild fortress bottles below gz-sim6, so I can separately update that to `python@3.12`.


scpeters:
build bottle

PR #2540, opened 2024-01-23
fortress: explicitly set python version

scpeters:
This sets the python version explicitly with the `Python3_EXECUTABLE` cmake variable. It also fixes the bindings installation location and adds a test.

The rebuild of gz-sim6 is part of #2527.


scpeters:
build bottle

scpeters:
~~~
==> brew audit osrf/simulation/ignition-math6 --online --git --skip-style
==> FAILED
osrf/simulation/ignition-math6
  * python modules have explicit framework links
    These python extension modules were linked directly to a Python
    framework binary. They should be linked with -undefined dynamic_lookup
    instead of -lpython or -framework Python.
      /usr/local/opt/ignition-math6/lib/python3.11/site-packages/ignition/math.cpython-311-darwin.so
Error: 1 problem in 1 formula detected.
~~~

I don't see this when building `gz-math7`, not sure what the difference is

scpeters:
I'll leave this as is for now

PR #2554, opened 2024-02-08
ignition-gazebo6 6.16.0

osrfbuild:


j-rivero:
release was a bump in linux revision, not need to do anything for brew

Issue #2504, opened 2024-01-04
Rebuild bottles after tinyxml2 10.0.0 is merged


scpeters:
tinyxml2 has had a new major release ([10.0.0](https://github.com/leethomason/tinyxml2/releases/tag/10.0.0)) and will require rebuilding several homebrew bottles when it is merged into homebrew-core. The following pull request was not merged but it may be resubmitted in the future.

* https://github.com/Homebrew/homebrew-core/pull/158645

Please supply your brew configuration:

### Brew configuration

~~~
$ brew config

~~~

### Brew linkage test result

[![Build Status](https://build.osrfoundation.org/buildStatus/icon?job=gz_common5-install_bottle-homebrew-amd64&build=60)](https://build.osrfoundation.org/view/gz-garden/job/gz_common5-install_bottle-homebrew-amd64/60/)

~~~
$ brew linkage --test gz-common5

~~~

### Checklist

- [x] Identify change in homebrew-core that broke the bottle https://github.com/Homebrew/homebrew-core/pull/158645
- [x] Identify list of affected formulae
    - dartsim@6.10.0
    - gazebo11
    - gz-common*
    - gz-fuel-tools*
    - gz-gui*
    - gz-launch*
    - gz-msgs*
    - gz-physics*
    - gz-sensors*
    - gz-sim*
    - gz-transport*
    - sdformat*
- [x] Remove broken bottles: for each affected formula, run `brew bump-revision --remove-bottle-block`
    - [X] https://github.com/osrf/homebrew-simulation/pull/2559
    - [x] https://github.com/osrf/homebrew-simulation/pull/2561
    - [x] https://github.com/osrf/homebrew-simulation/pull/2562
- [x] Rebuild broken bottles
    - [x] Harmonic
        - [x] https://github.com/osrf/homebrew-simulation/pull/2560
        - [x] https://github.com/osrf/homebrew-simulation/pull/2563
        - [x] https://github.com/osrf/homebrew-simulation/pull/2565
    - [x] Garden
        - [x] https://github.com/osrf/homebrew-simulation/pull/2560
        - [x] https://github.com/osrf/homebrew-simulation/pull/2564
        - [x] https://github.com/osrf/homebrew-simulation/pull/2566
    - [x] Fortress
        - [x] https://github.com/osrf/homebrew-simulation/pull/2567
    - [x] Citadel
        - [x] https://github.com/osrf/homebrew-simulation/pull/2568
[...]

PR #2568, opened 2024-02-14
Rebuild gazebo11 dependencies for tinyxml2

scpeters:
Part of #2504.

There are several dependencies shared by Citadel and Gazebo11. This rebuilds those common dependencies so that gazebo11 and the rest of Citadel can be built separately.

Pull request commits created with the following command:

~~~
for f in $(grep -L '^ *bottle do' $(grep -rnI 'depend.*"ignition-' gazebo11.rb | sed -e 's@.* "@@' -e 's@"@.rb@'))
do
     brew bump-revision --message="rebuild bottle" $f
done
~~~


scpeters:
build bottle

Issue #2580, opened 2024-02-23
Use python@3.12 for formulae with python bindings


scpeters:
Now that `python@3.12` is the default `python3` formula (see https://github.com/Homebrew/homebrew-core/pull/150390), we should update our formulae to use python@3.12 for the formulae with python bindings.

* [x] Ionic #2581
* [x] Harmonic
    * [x] https://github.com/osrf/homebrew-simulation/pull/2582
    * [x] https://github.com/osrf/homebrew-simulation/pull/2583
    * [x] gz-harmonic https://github.com/osrf/homebrew-simulation/pull/2726
* [x] Garden
    * [x] https://github.com/osrf/homebrew-simulation/pull/2582
    * [x] https://github.com/osrf/homebrew-simulation/pull/2583
    * [x] gz-garden https://github.com/osrf/homebrew-simulation/pull/2726

scpeters:
> Is it some relationship between this and [PX4/PX4-Autopilot#22826](https://github.com/PX4/PX4-Autopilot/issues/22826) ?

gazebo-classic doesn't use python bindings, so I don't think this is related to your PX4 issue

mengchaoheng:
> gazebo-classic doesn't use python bindings,

I mean, as a px4 user, when I use gazebo11 for simulation, it can work normally even if my python3 environment is not python3.12, right? @scpeters 

Sorry, I am currently troubled by px4 gazebo simulation problems. All compilation tasks are completed but gazebo fails.

scpeters:
> > gazebo-classic doesn't use python bindings,
> 
> I mean, as a px4 user, when I use gazebo11 for simulation, it can work normally even if my python3 environment is not python3.12, right? @scpeters
> 
> Sorry, I am currently troubled by px4 gazebo simulation problems. All compilation tasks are completed but gazebo fails.

gazebo11 does not directly use python. Unfortunately I can't answer questions about px4, so I recommend asking in a more relevant forum.

PR #2583, opened 2024-02-23
Garden, Harmonic: rebuild bottles with python 3.12

scpeters:
Part of https://github.com/osrf/homebrew-simulation/issues/2580. Follow-up to #2582.


scpeters:
there was a problem building the garden and harmonic bottles, so I rebased to remove those revision bumps

Issue #2587, opened 2024-03-11
abseil bump broke bottles


j-rivero:
The bump for abseil https://github.com/Homebrew/homebrew-core/commit/b794c6a136a70e7f15ab9002d8d228627f9a4d07 broke a good bunch of our bottles.

Let's try to rebuild the bottles:

 -  [x] Disabled broken bottles https://github.com/osrf/homebrew-simulation/pull/2588
 -  [ ] Rebuild the bottles
     - [x] Citadel https://github.com/osrf/homebrew-simulation/pull/2589
     - [x] Fortress https://github.com/osrf/homebrew-simulation/pull/2589
     - [x] Garden https://github.com/osrf/homebrew-simulation/pull/2591
     - [x] Harmonic https://github.com/osrf/homebrew-simulation/pull/2592
     - [x] gazebo11 https://github.com/osrf/homebrew-simulation/pull/2590

scpeters:
Under `Rebuild the bottles`, I added a checklist for each Gazebo collection and gazebo11

PR #2589, opened 2024-03-12
Citadel, Fortress: rebuild bottles for abseil (#2587)

j-rivero:
Part of #2587

### DIFF
--- a/Formula/ignition-fuel-tools4.rb
+++ b/Formula/ignition-fuel-tools4.rb
@@ -8,6 +8,12 @@ class IgnitionFuelTools4 < Formula
 
   head "https://github.com/gazebosim/gz-fuel-tools.git", branch: "ign-fuel-tools4"
 
+  bottle do
+    root_url "https://osrf-distributions.s3.amazonaws.com/bottles-simulation"
+    sha256 cellar: :any, ventura:  "f70d1702a8eb967e1af14d9bd2650fb3fd2329c92acb5d93c879c306c7e099d8"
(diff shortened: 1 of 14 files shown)


j-rivero:
@osrf-jenkins build bottle

j-rivero:
Oh I think that I probably need to add the block for bottles to actually build them :facepalm:

scpeters:
> Oh I think that I probably need to add the block for bottles to actually build them 🤦

no, it's enough to just bump the revision and trigger a build in a pull request

I don't see any issues with the `ignition-*` revision bumps, but the `gz-*` revision bumps are incomplete. There is some flexibility in choosing which formulae to bump at a given time, as long as the dependencies of each bumped formula are either already bottled or also receive a revision bump. For example, `gz-transport*` are bumped in this pull request, but `gz-msgs*` are not, so the `gz-transport*` bottles will be skipped. The `gz-sim*` bottles will be skipped for a similar reason.

If the build completes successfully, I predict that bottle blocks will be added to the `ignition-*` formulae but not to the `gz-*` formulae in this pull request. After it completes, the pedantic approach would be to rebase this branch and remove the commits bumping `gz-*`, and then merge the changes to `ignition-*` only, and bump `gz-*` in a separate pull request. It wouldn't break anything to just merge it as is with extra revision bumps, but it will keep the git history more tidy if the revision bumps without bottles were redacted.

The brew scripts have gotten better over time; it used to be that a build failure of a single package would cause all the builds to fail. In those times, I became accustomed to isolating changes to each dependency tree in single pull request. That seems to not be needed anymore.

PR #2636, opened 2024-04-12
gazebo11: rebuild for ffmpeg7

scpeters:
Part of https://github.com/osrf/homebrew-simulation/issues/2628.

PR #2637, opened 2024-04-12
Update gazebo11.rb to build against ffmpeg@6 instead of ffmpeg (v. 7)

gmerlino:
Fixes build failure.

### DIFF
--- a/Formula/gazebo11.rb
+++ b/Formula/gazebo11.rb
@@ -15,7 +15,7 @@ class Gazebo11 < Formula
   depends_on "bullet"
   depends_on "dartsim"
   depends_on "doxygen"
-  depends_on "ffmpeg"
+  depends_on "ffmpeg@6"
   depends_on "freeimage"
   depends_on "graphviz"
   depends_on "gts"


scpeters:
Thanks for the suggestion; I think this would work, but we'd have to set some environment variables in order to compile gazebo from source (see https://github.com/gazebo-tooling/release-tools/pull/636). Since we've already merged support for ffmpeg 7 in https://github.com/gazebosim/gz-common/pull/329, I'm inclined to port that patch to gazebo11

scpeters:
> Thanks for the suggestion; I think this would work, but we'd have to set some environment variables in order to compile gazebo from source (see [gazebo-tooling/release-tools#636](https://github.com/gazebo-tooling/release-tools/pull/636)). Since we've already merged support for ffmpeg 7 in [gazebosim/gz-common#329](https://github.com/gazebosim/gz-common/pull/329), I'm inclined to port that patch to gazebo11

I've attempted a patch in https://github.com/gazebosim/gazebo-classic/pull/3380

gmerlino:
> Thanks for the suggestion; I think this would work, but we'd have to set some environment variables in order to compile gazebo from source (see [gazebo-tooling/release-tools#636](https://github.com/gazebo-tooling/release-tools/pull/636)). Since we've already merged support for ffmpeg 7 in [gazebosim/gz-common#329](https://github.com/gazebosim/gz-common/pull/329), I'm inclined to port that patch to gazebo11

I see your point about the envs (PKG_CONFIG_PATH I reckon). Building upon install works without any additional changes to the formula, though.

scpeters:
Thanks for your suggestion; this is a valid approach, but I've chosen to rebuild with the patch from https://github.com/gazebosim/gazebo-classic/pull/3380 in https://github.com/osrf/homebrew-simulation/pull/2636, so I'm going to close this now

PR #2723, opened 2024-08-12
gz-math8: remove test based on manual compiler flags

azeey:
As reported in https://github.com/gazebosim/gz-math/pull/614#issuecomment-2284101515, the gz-math8 bottle install jobs are broken after https://github.com/gazebosim/gz-math/pull/614. This is because the PR made changes to the `SignalStats.hh` file to include a header from `gz-utils`. One solution is to update the manual flags to include `gz-utils3`, but that proved to be nontrivial because the variables defined by homebrew are specific to the package being built, in this case gz-math8. Figuring out the library and include directories for `gz-utils3` would involve using the `HOMEBREW_PREFIX` variable and somehow determining the version of `gz-utils3`. So, instead I opted to remove this test. @scpeters suggested implementing a replacement with `pkg-config`, but I think we should do that in follow up PR.


scpeters (approved):
I removed one more line in https://github.com/osrf/homebrew-simulation/pull/2723/commits/52c5b9d88cbd360caaaf9a073bcc6ba27515ce2d and now it passes for me

PR #2724, opened 2024-08-13
gz-math8: add pkg-config test

scpeters:
Follow-up to #2723. The test is adapted from that used in other formulae, like [sdformat14](https://github.com/osrf/homebrew-simulation/blob/master/Formula/sdformat14.rb#L74-L82).

### DIFF
--- a/Formula/gz-math8.rb
+++ b/Formula/gz-math8.rb
@@ -10,6 +10,7 @@ class GzMath8 < Formula
   depends_on "cmake" => :build
   depends_on "doxygen" => :build
   depends_on "pybind11" => :build
+  depends_on "pkg-config" => :test
   depends_on "eigen"
   depends_on "gz-cmake4"
   depends_on "gz-utils3"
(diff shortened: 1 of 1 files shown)


Formula/gz-math8.rb
```diff
@@ -52,6 +53,15 @@ def install
       add_executable(test_cmake test.cpp)
       target_link_libraries(test_cmake gz-math8::gz-math8)
     EOS
+    system "pkg-config", "gz-math8"
+    cflags = `pkg-config --cflags gz-math8`.split
+    system ENV.cc, "test.cpp",
+                   *cflags,
+                   "-L#{lib}",
+                   "-lgz-math8",
```
azeey, line 61: Can we get the linker flags from `pkg-config` as well?

scpeters: sure thing: https://github.com/osrf/homebrew-simulation/pull/2724/commits/6e19949b8f19770b466b62239f6ce5ab2165572a

the sdformat14 formula is only using the `cflags` from pkg-config, but I copied the approach for using linker flags from a gz-physics formula

azeey (approved):

PR #2732, opened 2024-08-22
gz-cmake4 4.0.0~pre1

osrfbuild:


azeey:
@scpeters if we build bottles here, does this mean that downstream libraries on `main` branches wil start using it instead of building from source? On Ubuntu, they will be using nightlies, on Windows, they'll be building from source, so I want to make sure we're not introducing a descrepancy.

scpeters:
this PR changes the url from a git branch to a prerelease tarball, so if we merge it then it will no longer build from the GitHub branch, regardless of whether it's bottled or not

so if we want to keep building from git branches instead of prerelease tarballs, then I would suggest that we close these prerelease pull requests until we are ready to build from a tarball

scpeters:
(also, your comment included the magic phrase that matches our regular expression in [brew_release.dsl](https://github.com/gazebo-tooling/release-tools/blob/7947017b7ce5ea73dff78f6a8701f7197d6344a4/jenkins-scripts/dsl/brew_release.dsl#L192), so it triggered an additional bottle builder job. it caused no harm in this case, just an FYI)

PR #2742, opened 2024-08-25
gazebo11: patch for boost 1.86

scpeters:
Patch from https://github.com/gazebosim/gazebo-classic/pull/3397 to fix build with boost 1.86.


scpeters:
build bottle

PR #2747, opened 2024-08-26
gz-common6: depend on spdlog

iche033:
to go with: https://github.com/gazebosim/gz-common/pull/615

note: `gz-utils3` should pull in `spdlog` so this may not be necessary.


scpeters (approved):
looks fine to me

scpeters:
> note: `gz-utils3` should pull in `spdlog` so this may not be necessary.

to clarify, this isn't explicitly needed, because `gz-utils3` is pulling it in, but brew is happier when we explicitly depend on libraries that we link against, so I approve merging it

PR #2773, opened 2024-08-29
Ionic prereleases

azeey:
Pulls changes from

- https://github.com/osrf/homebrew-simulation/pull/2732
- https://github.com/osrf/homebrew-simulation/pull/2764
- https://github.com/osrf/homebrew-simulation/pull/2740
- https://github.com/osrf/homebrew-simulation/pull/2751
- https://github.com/osrf/homebrew-simulation/pull/2765
- https://github.com/osrf/homebrew-simulation/pull/2750
- https://github.com/osrf/homebrew-simulation/pull/2760
- https://github.com/osrf/homebrew-simulation/pull/2766
- https://github.com/osrf/homebrew-simulation/pull/2761
- https://github.com/osrf/homebrew-simulation/pull/2767
- https://github.com/osrf/homebrew-simulation/pull/2753
- https://github.com/osrf/homebrew-simulation/pull/2769
- https://github.com/osrf/homebrew-simulation/pull/2768
- https://github.com/osrf/homebrew-simulation/pull/2770
- https://github.com/osrf/homebrew-simulation/pull/2771
- https://github.com/osrf/homebrew-simulation/pull/2772

PR #2774, opened 2024-08-29
gz-common6: disable pkg-config test

azeey:
bottles failed to build for gz-common6 in #2773 due to a test failure.


azeey:
build bottle

scpeters:
build bottle

scpeters:
the cmake test works, but the pkg-config test is broken; I don't see `fmt` listed as a pkg-config dependency; I'm guessing that would need to be fixed in order to re-enable it

scpeters:
build bottle

Issue #2745, opened 2024-08-26
Fix bottles broken by fmt 11.0.2


scpeters:
I'm guessing that `fmt` 11.0.2 (used by dartsim) that will be upgraded in https://github.com/Homebrew/homebrew-core/pull/178811 may break some of our bottles.

Please supply your brew configuration:

### Brew configuration

~~~
$ brew config

~~~

### Brew linkage test result

~~~
$ brew linkage --test gz-common5

~~~

### Checklist

- [x] Identify change in homebrew-core that broke the bottle https://github.com/Homebrew/homebrew-core/pull/178811
- [x] Identify list of affected formulae
    - [x] gz-physics5
    - [x] gz-physics6
    - [x] gz-physics7
    - [x] gazebo11
- [x] Remove broken bottles: for each affected formula, run `brew bump-revision --remove-bottle-block`
    - [x] https://github.com/osrf/homebrew-simulation/pull/2775
    - [x] remove gazebo11 bottle https://github.com/osrf/homebrew-simulation/pull/2777
- [x] Rebuild broken bottles
    - [x] https://github.com/osrf/homebrew-simulation/pull/2776
    - [x] gazebo11 https://github.com/osrf/homebrew-simulation/pull/2779

scpeters:
it looks like the gz-physics bottles for fortress, garden and harmonic are broken due to an update to `fmt`, which is used by `dartsim`. we should remove those bottles and then rebuild them

scpeters:
I forgot about gazebo11; that bottle needs to be removed as well

PR #2775, opened 2024-08-29
gz-physics*: remove bottles for fmt

scpeters:
A recent update to fmt has broken several gz-physics bottles, via its usage by dartsim. I've added an explicit dependency on fmt to help with identifying future issues like this.

Part of #2745.


Formula/ignition-physics5.rb
```diff
@@ -4,22 +4,17 @@ class IgnitionPhysics5 < Formula
...
-    sha256 cellar: :any, monterey: "debb833c7c0ef7f22bc3b33cab2d828e442905efa9c09ea02394d4316e21d84e"
-  end
-
   depends_on "cmake" => :build
 
   depends_on "gz-plugin2" => :test
 
   depends_on "bullet"
   depends_on "dartsim"
+  depends_on "fmt"
```
azeey, line 17: Is this now necessary for some reason?

scpeters: it's an implicit dependency via `dartsim`. brew gives some warnings about linking to formula without an explicit dependency, so this reduces those warnings

azeey (approved):

PR #2780, opened 2024-09-03
gz-gui9: disable pkg-config test

scpeters:
Follow up to #2774.

Test is currently failing, so disable it.

[Build Status](https://build.osrfoundation.org/view/gz-ionic/job/gz_gui9-install_bottle-homebrew-amd64/267/) https://build.osrfoundation.org/view/gz-ionic/job/gz_gui9-install_bottle-homebrew-amd64/267/

### DIFF
--- a/Formula/gz-gui9.rb
+++ b/Formula/gz-gui9.rb
@@ -101,16 +101,16 @@ class GzGui9 < Formula
       add_executable(test_cmake test.cpp)
       target_link_libraries(test_cmake gz-gui9::gz-gui9)
     EOS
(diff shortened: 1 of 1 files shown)


azeey (approved):
For some reason `pkg-config --libs` is not including `gz-utils3-logs` and its dependencies. This is true in Linux as well. I'm looking to see if it's an issue in gz-cmake.

azeey:
It looks like gz-cmake's generated pkg-config file does not included required components from dependencies. https://github.com/gazebosim/gz-cmake/issues/450

PR #2798, opened 2024-09-22
Build sonoma bottles for gz-cmake2, dependents

scpeters:
Start with formulae shared between citadel, fortress.


scpeters:
build bottle
brew-bot-tag: build-for-new-distro-sonoma

PR #2799, opened 2024-09-22
Citadel: build sonoma bottles part 2

scpeters:
Make trivial changes to allow building all formulae used in citadel and gazebo11 that don't use protobuf, since that will be addressed by #2791.

PR #2803, opened 2024-09-23
gz-rendering3: fix rpath issues

scpeters:
This formula's Sonoma bottle failed its `brew test` in #2799. I noticed some rpath fixes in the fortress formula that hadn't been backported to citadel, so I'm backporting the fix here and bumping the revision to that the Ventura bottle will be rebuilt as well.


scpeters:
build bottle

PR #2814, opened 2024-09-26
Rebuild citadel bottles for protobuf 28.2

scpeters:
Part of #2791.

Created using:

~~~
for f in $(.github/ci/unbottled_dependencies.sh gz-citadel)
do
    brew bump-revision --message="rebuild for protobuf" $f
done
~~~

PR #2828, opened 2024-09-30
gz-gui3: fix rpath for plugins and rebuild

scpeters:
Follow up to #2814, similar to https://github.com/osrf/homebrew-simulation/pull/2803.


scpeters:
build bottle

scpeters:
sonoma bottle was successful this time

Issue #2791, opened 2024-09-15
Rebuild bottles broken by protobuf 28.2


scpeters:
[![Build Status](https://build.osrfoundation.org/buildStatus/icon?job=gz_msgs11-install_bottle-homebrew-amd64&build=274)](https://build.osrfoundation.org/view/gz-ionic/job/gz_msgs11-install_bottle-homebrew-amd64/274/) https://build.osrfoundation.org/view/gz-ionic/job/gz_msgs11-install_bottle-homebrew-amd64/274/

### Brew configuration

Please supply your brew configuration:

~~~
+ brew config
HOMEBREW_VERSION: 4.3.21
ORIGIN: https://github.com/Homebrew/brew
HEAD: c5d09b411776547bdbc8e0435deaf690ca632413
Last commit: 3 days ago
Core tap HEAD: 90a452aeb33fd7ffe5f9da10aa2a6b8df5bdcca5
Core tap last commit: 14 minutes ago
Core tap JSON: 14 Aug 11:15 UTC
Core cask tap HEAD: c80f1c957111533b5698f8200a0af0c3c464fcf6
Core cask tap last commit: 56 minutes ago
Core cask tap JSON: 14 Aug 11:15 UTC
HOMEBREW_PREFIX: /usr/local
HOMEBREW_CASK_OPTS: []
HOMEBREW_GIT_EMAIL: osrfbuild@osrfoundation.org
HOMEBREW_GIT_NAME: OSRF Build Bot
HOMEBREW_MAKE_JOBS: 12
HOMEBREW_NO_AUTO_UPDATE: set
HOMEBREW_NO_INSTALL_FROM_API: set
HOMEBREW_SORBET_RUNTIME: set
HOMEBREW_UPDATE_TO_TAG: set
Homebrew Ruby: 3.3.4 => /usr/local/Homebrew/Library/Homebrew/vendor/portable-ruby/3.3.4_1/bin/ruby
CPU: dodeca-core 64-bit kabylake
Clang: 14.0.0 build 1400
Git: 2.46.1 => /usr/local/bin/git
Curl: 7.87.0 => /usr/bin/curl
macOS: 12.6.6-x86_64
CLT: 14.2.0.0.1.1668646533
Xcode: N/A
~~~

### Brew linkage test result

~~~
+ brew linkage --test gz-msgs11
Broken dependencies:
  /usr/local/opt/protobuf/lib/libprotobuf.28.0.0.dylib (protobuf)
[...]

PR #2830, opened 2024-10-01
gz-sim3, gz-launch2: build sonoma bottles

scpeters:
Part of #2791, follow-up to #2828.


scpeters:
build bottle
brew-bot-tag: build-for-new-distro-sonoma

PR #2843, opened 2024-11-01
gz-msgs10: support multiple python versions

scpeters:
Part of https://github.com/osrf/homebrew-simulation/issues/2834.

This package doesn't generate python binaries, just python files, so we can just symlink the generated python files to versioned python folders.


scpeters:
build bottle

PR #2856, opened 2024-11-08
gz-sim8 8.7.0

osrfbuild:


scpeters:
I'm going to merge this without a bottle since there are dependencies that need to be built first

PR #2905, opened 2024-12-02
gz-sim8: use python@3.13

scpeters:
Part of https://github.com/osrf/homebrew-simulation/issues/2834. For now just update to use the current default python version.


scpeters:
build bottle

Issue #2916, opened 2024-12-17
Rebuild bottles broken by boost 1.87.0


scpeters:
Please supply your brew configuration:

### Brew configuration

~~~
+ brew config
HOMEBREW_VERSION: 4.4.12
ORIGIN: https://github.com/Homebrew/brew
HEAD: 3001afe88112c13b5aafdd92e097680897060881
Last commit: 32 hours ago
Branch: stable
Core tap HEAD: 27daa063da9d84512dd4134f6982cd113b49b7c0
Core tap last commit: 5 minutes ago
Core tap JSON: 12 Dec 11:13 UTC
Core cask tap HEAD: 3fa90d3483a87f42b36d2277702a34c238bd43ac
Core cask tap last commit: 34 minutes ago
Core cask tap JSON: 12 Dec 11:13 UTC
HOMEBREW_PREFIX: /usr/local
HOMEBREW_CASK_OPTS: []
HOMEBREW_GIT_EMAIL: osrfbuild@osrfoundation.org
HOMEBREW_GIT_NAME: OSRF Build Bot
HOMEBREW_MAKE_JOBS: 12
HOMEBREW_NO_AUTO_UPDATE: set
HOMEBREW_NO_INSTALL_FROM_API: set
HOMEBREW_SORBET_RUNTIME: set
HOMEBREW_UPDATE_TO_TAG: set
Homebrew Ruby: 3.3.6 => /usr/local/Homebrew/Library/Homebrew/vendor/portable-ruby/3.3.6/bin/ruby
CPU: dodeca-core 64-bit kabylake
Clang: 15.0.0 build 1500
Git: 2.47.1 => /usr/local/bin/git
Curl: 8.6.0 => /usr/bin/curl
macOS: 14.5-x86_64
CLT: 15.3.0.0.1.1708646388
Xcode: N/A

~~~

### Brew linkage test result

[![Build Status](https://build.osrfoundation.org/buildStatus/icon?job=gz_physics2-install_bottle-homebrew-amd64&build=366)](https://build.osrfoundation.org/view/gz-citadel/job/gz_physics2-install_bottle-homebrew-amd64/366/) https://build.osrfoundation.org/view/gz-citadel/job/gz_physics2-install_bottle-homebrew-amd64/366/

~~~
+ brew linkage --test gz-physics2
Broken dependencies:
  /usr/local/opt/boost/lib/libboost_atomic-mt.dylib (boost)
[...]

PR #2935, opened 2025-01-11
gazebo11: use boost 1.85

scpeters:
This is a workaround for #2916 and https://github.com/gazebosim/gazebo-classic/issues/3414 that uses forked versions of `boost` and `ogre1.9`:

* `boost@1.85.0` (derived using `brew extract bost@1.85 osrf/simulation`)
* `ogre1.9-with-boost1.85` (copied from `ogre1.9`)

If we find a fix for https://github.com/gazebosim/gazebo-classic/issues/3414, we can revert to using the system version of `boost`.

Note that `boost@1.85.0` and `ogre1.9-with-boost1.85` are keg-only, which means they are not sym-linked to `HOMEBREW_PREFIX`, so compiling gazebo11 from source or compiling custom plugins will require setting additional environment variables, similar to the approach for `qt@5` in the [build-from-source documentation](https://gazebosim.org/docs/latest/install_osx_src/#install-dependencies).

### DIFF
--- a/Formula/boost@1.85.0.rb
+++ b/Formula/boost@1.85.0.rb
@@ -1,0 +1,121 @@
+class BoostAT1850 < Formula
+  desc "Collection of portable C++ source libraries"
+  homepage "https://www.boost.org/"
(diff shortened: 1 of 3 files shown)


scpeters:
build bottle

scpeters:
build bottle

PR #2956, opened 2025-02-05
gz-common5 5.7.1

osrfbuild:


scpeters:
cherry-picked into #2957

PR #2985, opened 2025-02-20
gz-sim10: depend on libwebsockets

iche033:
Related issue: https://github.com/gazebosim/gz-launch/issues/288
Needed by: https://github.com/gazebosim/gz-sim/pull/2792

The websocket server is being ported from gz-launch to gz-sim. This adds `libwebsockets` dependency to gz-sim10 for Jetty. Note: I don't see this in the gz-launch9 formula as I think it's only an optional dependency there.

### DIFF
--- a/Formula/gz-sim10.rb
+++ b/Formula/gz-sim10.rb
@@ -27,6 +27,7 @@ class GzSim10 < Formula
   depends_on "gz-tools2"
   depends_on "gz-transport14"
   depends_on "gz-utils3"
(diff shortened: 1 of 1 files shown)


iche033:
Not sure why brew test-bot started producing warnings for other formulas. Here's a PR to fix the warnings: https://github.com/osrf/homebrew-simulation/pull/2986

scpeters (approved):

PR #2989, opened 2025-03-03
fix: resolve deprecation warning

ddalton15:
### Overview:

- Resolve warning here:
```Warning: Calling inreplace(paths, before, after, false) is deprecated! Use inreplace(paths, before, after, audit_result: false) instead.
Please report this issue to the osrf/homebrew-simulation tap (not Homebrew/brew or Homebrew/homebrew-core), or even better, submit a PR to fix it:
  /opt/homebrew/Library/Taps/osrf/homebrew-simulation/Formula/simbody.rb:3```


scpeters (approved):
thanks for the fix!

PR #3055, opened 2025-05-22
sdformat15: fix standalone executable rpath and rebuild

scpeters:
Fixes rpath for standalone executable and adds test.


scpeters:
build bottle

PR #3089, opened 2025-07-28
Workflow: use ubuntu-latest

scpeters:
See if this fixes linter error.


scpeters:
test bot workflow seems fixed from no action on my part, closing

PR #3287, opened 2025-12-19
Rotary collection: gz-cmake, gz-utils

scpeters:
Prototype of `gz-rotary-*` formulae, inspired by the Rotary release repos proposal in https://github.com/gazebo-tooling/release-tools/issues/1446. The goal of this prototype is to show how `gz-rotary-*` formulae could be added that would always build against each other from the `main` branch and would require minimal changes when creating stable branches or releases. As such, I propose that all formula names are all prefixed with `gz-rotary-` (even `gz-rotary-sdformat`). This PR implements `gz-rotary-cmake` and `gz-rotary-utils` that install from main branch and conflict with jetty versions.

* When new formulae are created, conflicts need to be added to the `gz-rotary-*` formulae, similiar to the `gz-jetty-*` conflicts in this prototype.
* Currently our [brew CI uses package major version number](https://github.com/gazebo-tooling/release-tools/blob/a11e496f9548bd4ee0b1fa0b6b91396a8722ad57/jenkins-scripts/lib/project-default-devel-homebrew-amd64.bash#L12-L21) to install dependencies, so we maintain numbered aliases that point to `gz-rotary-*` formulae.
* To avoid requiring an explicity `version` tag, this prototype uses only the `head` specification, which requires installing with `--HEAD` (see https://github.com/gazebo-tooling/release-tools/pull/1451). The alternative is to use `url` instead of `head` and explicitly specify the `version` tag (like [here](https://github.com/osrf/homebrew-simulation/blob/8c659dcfe9e1edf2e99e89942016213b11c52da1/Formula/gz-cmake5.rb#L4-L5)).

### DIFF
--- a/Formula/gz-fuel-tools10.rb
+++ b/Formula/gz-fuel-tools10.rb
@@ -4,16 +4,9 @@ class GzFuelTools10 < Formula
   url "https://osrf-distributions.s3.amazonaws.com/gz-fuel-tools/releases/gz-fuel_tools-10.1.0.tar.bz2"
   sha256 "37ae351be9a9b281d078e36068422dd096f59f46c72c4ef490800dfeb7653e1d"
   license "Apache-2.0"
-  revision 28
+  revision 29
 
   head "https://github.com/gazebosim/gz-fuel-tools.git", branch: "gz-fuel-tools10"
-
-  bottle do
-    root_url "https://osrf-distributions.s3.amazonaws.com/bottles-simulation"
-    sha256 cellar: :any, arm64_sequoia: "a5638619d582bf9638d52415fcc0e57627d21dc8bed5b82315e85ef5177289d1"
-    sha256 cellar: :any, arm64_sonoma:  "dc187e95efcb95847d2851774276b1ad796f2d60fede0795d15ab58734a6b8ce"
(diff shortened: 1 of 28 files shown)


Formula/gz-rotary-cmake.rb
```diff
@@ -0,0 +1,50 @@
...
+  desc "CMake helper functions for building robotic applications"
+  homepage "https://gazebosim.org"
+  license "Apache-2.0"
+
+  head "https://github.com/gazebosim/gz-cmake.git", branch: "main"
+
+  depends_on "cmake"
+  depends_on "pkgconf"
+
+  conflicts_with "gz-jetty-cmake", because: "both install gz-cmake"
```
azeey: Do we also need a conflict with `gz-cmake5` here?

azeey (commented):
LGTM. Just one question

Formula/gz-rotary-cmake.rb
azeey: Oh, I was just checking if it was missed. If using just the collection aliases is sufficient, then by all means.

scpeters:
> opened a draft PR for updating our CI scripts to install with `--HEAD`
> 
> would also need to update macos github workflow scripts in each repo that has one

PR is ready for review:

* https://github.com/gazebo-tooling/release-tools/pull/1451

azeey (approved):

PR #3288, opened 2026-01-02
Add aliases for versions post Jetty

azeey:
Part of https://github.com/gazebosim/gz-jetty/issues/70


scpeters:
I have an alternate PR that adds the needed aliases in https://github.com/osrf/homebrew-simulation/pull/3287; what do you think?

j-rivero:
https://github.com/osrf/homebrew-simulation/pull/3361 is a superset of this one

Issue #3298, opened 2026-01-12
freeimage has been deprecated


scpeters:
The freeimage formula has been deprecated in https://github.com/Homebrew/homebrew-core/pull/262236 since it has multiple CVEs.

scpeters:
we have dropped `freeimage` from the `rotary` collection on macOS as of #3394

scpeters:
dropped `freeimage` from gz-common7 in https://github.com/osrf/homebrew-simulation/pull/3554

for Jetty, we would need to make the following changes to gz-rendering10:

* drop ogre1.9
* change the ogre2.3 formula to what is currently in [Formula/gz-rotary-ogre2.3-vendor.rb](https://github.com/osrf/homebrew-simulation/blob/master/Formula/gz-rotary-ogre2.3-vendor.rb)

I don't think we would backport the switch from freeimage to STB any further, so I expect that we would drop support for Ionic and earlier when freeimage is [disabled in January 2027](https://github.com/Homebrew/homebrew-core/blob/main/Formula/f/freeimage.rb#L28-L31)

scpeters:
> ## Migration plan
> * [x] Add `gz-rotary-ogre2.3-vendor` formula without freeimage
>     * [x]  #3391 
>     * [x] #3408 
> * [x] Drop freeimage from Rotary https://github.com/osrf/homebrew-simulation/pull/3394 
> * [ ] Rename current `ogre2.3` formula to `ogre2.3-with-freeimage`
> * [ ] Rename `gz-rotary-ogre2.3-vendor` to `ogre2.3`
> * [ ] Jetty
>     * [ ] Use `ogre2.3` without freeimage
>     * [ ] Drop `ogre1.9` (wait until Jan 2027)


I added a migration plan to the PR description

PR #3394, opened 2026-03-20
gz-rotary-rendering: drop freeimage

scpeters:
* Switch from ogre2.3 to gz-rotary-ogre2.3-vendor
* Drop ogre1.9

Part of https://github.com/osrf/homebrew-simulation/issues/3298.

### DIFF
--- a/.github/ci/print-homebrew-simulation-badges.bash
+++ b/.github/ci/print-homebrew-simulation-badges.bash
@@ -1,0 +1,128 @@
+#!/bin/bash
+
+RELEASE_TOOLS_REPO="https://github.com/gazebo-tooling/release-tools"
+generated_jobs="$(mktemp -d)/generated_jobs.txt"
+curl --location "${RELEASE_TOOLS_REPO}/raw/master/jenkins-scripts/dsl/logs/generated_jobs.txt" \
+  --output "${generated_jobs}"
+
+JOB_URL='https://build.osrfoundation.org/job/'
+BADGE_URL='https://build.osrfoundation.org/buildStatus/icon?job='
+arches=( \
+  amd64 \
+  arm64 \
+)
+collections=( \
+  fortress \
+  harmonic \
+  ionic \
+  jetty \
+  rotary \
+)
+packages=( \
+  gz-cmake \
+  gz-common \
+  gz-fuel-tools \
+  gz-gui \
+  gz-launch \
+  gz-math \
+  gz-msgs \
+  gz-physics \
+  gz-plugin \
+  gz-rendering \
+  gz-sensors \
+  gz-sim \
+  gz-tools \
+  gz-transport \
+  gz-utils \
+  sdformat \
+)
+
+# print table of badges
+# one row for each package / arch
+# one column for each collection
+# Status        | Arch | Fortress | Harmonic | Ionic | Jetty
+echo -n "Status        | Arch"
+for c in ${collections[@]};
+do
+  echo -n " | $c"
+done
+echo
+# ------------- | ---- | -------- | -------- | ----- | -------
+echo -n "------------- | ----"
+for c in ${collections[@]};
+do
+  echo -n " | $(echo $c | tr '[:print:]' '-')"
+done
+echo
+for p in ${packages[@]};
+do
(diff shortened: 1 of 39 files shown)


scpeters:
> There is an issue with RGB-to-RGBA conversion in ogre-next v2.3 when using STB. I've created [OGRECave/ogre-next#567](https://github.com/OGRECave/ogre-next/pull/567) to fix the issue and have tested locally.

thanks! I've approved the upstream PR and am adding the patch to `gz-rotary-ogre2.3-vendor` in #3408

scpeters:
> thanks! I've approved the upstream PR and am adding the patch to `gz-rotary-ogre2.3-vendor` in #3408

I added the patch to gz-rotary-ogre2.3-vendor in #3408 and marked this ready for review

taylorhoward92 (approved):
LGTM

scpeters:
we agreed at [Monday's Gazebo meeting](https://docs.google.com/document/d/1RpVMkwiKqXKnDjqtkviz4qp73oehivf4Z1x-fTITgZY/edit?tab=t.0#heading=h.n2fe63pha7wa) to fully drop freeimage on macOS on the main branch by dropping support for ogre 1.9

so I think we are ready to merge this. I will wait a few hours in case any Gazebo maintainers want to officially mark their approval

azeey (approved):
LGTM!

However, I noticed that gz-rotary-ogre2.3-vendor depends on `gz-jetty-plugin` https://github.com/osrf/homebrew-simulation/blob/39e3b5ffbfa93061d64cb91fac1361bb4c5318ca/Formula/gz-rotary-ogre2.3-vendor.rb#L19

Is that intentional? 

Also, if freeimage is going to be dropped altogether, shouldn't we be patching the ogre2.3 formula itself instead of creating gz-rotary-ogre2.3?

scpeters:
I'll answer these in reverse order

> Also, if freeimage is going to be dropped altogether, shouldn't we be patching the ogre2.3 formula itself instead of creating gz-rotary-ogre2.3?

I created the `gz-rotary-ogre2.3-vendor` formula so that we could experiment with an alternative ogre build without affecting existing stable releases like Jetty. Hopefully we can backport support for STB to Jetty and update the `ogre2.3` formula and drop `gz-rotary-ogre2.3-vendor` (at least until we want to test updates to the ogre formula)

> However, I noticed that gz-rotary-ogre2.3-vendor depends on `gz-jetty-plugin`
> 
> https://github.com/osrf/homebrew-simulation/blob/39e3b5ffbfa93061d64cb91fac1361bb4c5318ca/Formula/gz-rotary-ogre2.3-vendor.rb#L19
> 
> Is that intentional?

yes it does currently declare a `test` dependency on `gz-jetty-plugin`, and it is intentional because I want `gz-rotary-ogre2.3-vendor` to be bottled, which requires all dependencies (even test dependencies) to be bottled

more context:

* I especially want `gz-rotary-ogre2.3-vendor` to be vendored because ogre will opportunistically find `freeimage` if you have it installed on your system, which I experienced with local testing. Bottles are built in a controlled environment, which ensures that it is truly built without `freeimage`
* We started using `gz-plugin` in the `ogre2.3` in formula test to confirm that rpath was properly configured (see https://github.com/osrf/homebrew-simulation/commit/9fb0d087d7957f3fc79ccfcf208f39f95b34e5fa). When adding `gz-rotary-ogre2.3-vendor` in #3391, I initially built the bottles while depending on `gz-plugin2` and then switched to `gz-rotary-plugin` in https://github.com/osrf/homebrew-simulation/pull/3391/changes/f21fe4d18a6b39398e49683f026240bf0f49495b after the bottles were built. When adding another patch to ogre2.3 in #3408 to fix the color swapping, my initial attempt to rebuild the bottles failed since `gz-rotary-plugin` is not bottled, so I switched back to `gz-jetty-plugin` and just left it in place. As a test dependency it doesn't get installed unless you explicitly `brew install --include-test gz-rotary-ogre2.3-vendor`. If we really don't like it, we can switch it back to `gz-rotary-plugin`, but I think it doesn't hurt anything so I left it in place

azeey:
Makes sense. Thanks for the explanation!

PR #3425, opened 2026-05-01
gz-rotary-ogre2.3-vendor: remove tbb dependency

scpeters:
Threading is not used as we are not explicitly enabling it (see https://github.com/OGRECave/ogre-next/issues/173#issuecomment-764765319), so remove this unused dependency.


scpeters:
build bottle

PR #3495, opened 2026-07-16
gz-cmake4: add patch for cmake 4.4 warnings

scpeters:
Part of https://github.com/gazebosim/gz-cmake/issues/570.


scpeters:
build bottle

PR #3566, opened 2026-09-09
Rename ogre2.3 to ogre2.3-with-freeimage

scpeters:
The `freeimage` formula is deprecated and will be removed from homebrew-core in Jan 2027. Per the migration plan in #3298, the `gz-rotary-ogre2.3-vendor` formula that does not use freeimage has been added and tested in Rotary. That formula will be renamed to `ogre2.3`, and as the first step in that process, the current `ogre2.3` formula that depends on `freeimage` is being renamed to `ogre2.3-with-freeimage`.

All bottles that depend on `ogre2.3` will be invalidated by this formula name change, so this builds on bottle removal commits that were generated in #3565 by https://build.osrfoundation.org/job/generic-release-homebrew_remove_dependent_bottles/46/

### testing

* [Build Status](https://build.osrfoundation.org/job/gz_rendering10-install_bottle-homebrew-arm64/362/) https://build.osrfoundation.org/job/gz_rendering10-install_bottle-homebrew-arm64/362/
* [Build Status](https://build.osrfoundation.org/job/gz_sensors10-install_bottle-homebrew-arm64/366/) https://build.osrfoundation.org/job/gz_sensors10-install_bottle-homebrew-arm64/366/

PR #3567, opened 2026-09-09
Rename gz-rotary-ogre2.3-vendor to ogre2.3

scpeters:
This is a follow-up to https://github.com/osrf/homebrew-simulation/pull/3566 that completes the formula name change from `gz-rotary-ogre2.3-vendor` to `ogre2.3` as part of #3298.

Bottles will be rebuilt in subsequent pull requests.

### testing

* [Build Status](https://build.osrfoundation.org/job/gz_rotary_rendering-install_formula-homebrew-arm64/181/) https://build.osrfoundation.org/job/gz_rotary_rendering-install_formula-homebrew-arm64/181/

### DIFF
--- a/Aliases/gz-jetty-ogre2.3
+++ b/Aliases/gz-jetty-ogre2.3
@@ -1,1 +1,1 @@
-../Formula/ogre2.3.rb
+../Formula/ogre2.3-with-freeimage.rb
--- a/Formula/gz-gui10.rb
+++ b/Formula/gz-gui10.rb
@@ -4,15 +4,9 @@ class GzGui10 < Formula
   url "https://osrf-distributions.s3.amazonaws.com/gz-gui/releases/gz-gui-10.1.0.tar.bz2"
   sha256 "2bb422b974a783e6fbb8a57c5f0f2d4f7f9802ec3bb2b6766c7b07f00a193ed7"
   license "Apache-2.0"
-  revision 4
+  revision 5
 
   head "https://github.com/gazebosim/gz-gui.git", branch: "gz-gui10"
-
-  bottle do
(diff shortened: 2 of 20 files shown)


j-rivero:
Since we are changing the name of gz-rotary-ogre2.3-vendor, the package managers usually provide some mechanism to handle these transitions in renames so an upgrade knows what to do with the old package. Seems like brew provides something in the form of a json file that we can put somewhere https://docs.brew.sh/Rename-A-Formula , something like  `{"gz-rotary-ogre2.3-vendor": "ogre2.3"}` ?

scpeters:
> Since we are changing the name of gz-rotary-ogre2.3-vendor, the package managers usually provide some mechanism to handle these transitions in renames so an upgrade knows what to do with the old package. Seems like brew provides something in the form of a json file that we can put somewhere https://docs.brew.sh/Rename-A-Formula , something like `{"gz-rotary-ogre2.3-vendor": "ogre2.3"}` ?

we used to have this file but removed it in #2867 when old files were removed. I'll try adding it or using an alias

scpeters:
> > Since we are changing the name of gz-rotary-ogre2.3-vendor, the package managers usually provide some mechanism to handle these transitions in renames so an upgrade knows what to do with the old package. Seems like brew provides something in the form of a json file that we can put somewhere https://docs.brew.sh/Rename-A-Formula , something like `{"gz-rotary-ogre2.3-vendor": "ogre2.3"}` ?
> 
> we used to have this file but removed it in #2867 when old files were removed. I'll try adding it or using an alias

added in https://github.com/osrf/homebrew-simulation/pull/3567/commits/05508bd45981ed7faccf31970fc445c645ce7ca6

